PR #1, opened 2017-11-13
mi fix servlet dependency

madeye-matt:


pom.xml
```diff
@@ -183,6 +183,12 @@
     <dependency>
       <groupId>net.sourceforge.cobertura</groupId>
```
benjefferies: I guess it could be `<scope>test</scope>` too

PR #5, opened 2017-11-22
Docker java debug

madeye-matt:
Open port 5151 on docker container for debugging


pom.xml
```diff
@@ -255,6 +255,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-failsafe-plugin</artifactId>
+        <version>2.20.1</version>
```
benjefferies: Nice. I've created a story around this https://trello.com/c/135d76iw/4-fix-failsafe-plugin-warnings. Maybe we should add it to the parent pom?

PR #12, opened 2018-01-12
Don't fail on new fields in service responses

madeye-matt:
Also, updated version of rm-collectionexercisesvc-api to latest

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
     <dependency>
       <groupId>uk.gov.ons.ctp.product</groupId>
       <artifactId>collectionexercisesvc-api</artifactId>
-      <version>10.49.4-SNAPSHOT</version>
+      <version>10.49.4</version>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -49,7 +49,7 @@
     <dependency>
       <groupId>uk.gov.ons.ctp.product</groupId>
       <artifactId>collectionexercisesvc-api</artifactId>
-      <version>10.49.0</version>
+      <version>10.49.4-SNAPSHOT</version>
```
benjefferies: Assuming this will change to the release version

src/main/java/uk/gov/ons/ctp/response/action/ActionSvcApplication.java
```diff
@@ -219,7 +221,11 @@ public RestExceptionHandler restExceptionHandler() {
    */
   @Bean @Primary
   public CustomObjectMapper customObjectMapper() {
-    return new CustomObjectMapper();
+    CustomObjectMapper mapper = new CustomObjectMapper();
+
+    mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
benjefferies: I agree this is the right thing to be doing

PR #17, opened 2018-02-05
Update api.md

gemmai95:
Update api.md to bring in line with other services.


gemmai95:
[Update Api.md for java services](https://trello.com/c/JawttNv0/50-update-apimd-for-java-services)

jcox-dev:
In technical documentation we should be using "ID" (originally an acronym for "Identity Document" but now used as an abbr. for "identifier" or "identity") as "id" is a separate English word of its own. I'm not sure we can get away with the uncommon "Id" here, outside of code-space.

davidmort:
@jcox-dev looks like your comment has been addressed, is this good to merge now?

PR #24, opened 2018-03-09
Added schema file

madeye-matt:
To allow the service to spin up from an empty database

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -59,6 +59,7 @@ spring:
     username: actionsvc
     password: actionsvc
     driverClassName: org.postgresql.Driver
+    continue-on-error: true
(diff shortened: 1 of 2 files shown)


src/main/resources/application.yml
```diff
@@ -59,6 +59,7 @@ spring:
     username: actionsvc
     password: actionsvc
     driverClassName: org.postgresql.Driver
+    continue-on-error: true
```
tej99: Is this something we want in all environments? Would it not be better to add this in our ras-rm-docker-dev configuration instead?

benjefferies: I think it only applies to the local environment because it's not been added in the application-cloud.yml

PR #25, opened 2018-03-13
Use new endpoint to get parties with associations filtered by survey

jwgwalton:


jwgwalton:
@davidmort @pricem14pc I couldn't find a release 16 tagged branch to hang this pull request off

PR #27, opened 2018-03-26
Refactor action service

benjefferies:
Attempt to make some of the involved large methods smaller and more
readable so that it will be easier to maintain and improve/change going
forward. A prerequisite before looking to implement
https://trello.com/c/qPlEbzaW/68-investigate-use-of-select-for-update-skip-locked-in-place-of-the-current-redis-locking-mechanism-when-using-postgres-as-a-queue

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
@@ -15,13 +15,13 @@ import uk.gov.ons.ctp.response.action.domain.model.Action;
 import uk.gov.ons.ctp.response.action.domain.model.ActionType;
 import uk.gov.ons.ctp.response.action.domain.repository.ActionRepository;
 import uk.gov.ons.ctp.response.action.domain.repository.ActionTypeRepository;
-import uk.gov.ons.ctp.response.action.representation.ActionDTO;
 import uk.gov.ons.ctp.response.action.representation.ActionDTO.ActionState;
 import uk.gov.ons.ctp.response.action.service.ActionProcessingService;
 
 import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 /**
@@ -34,7 +34,7 @@ import java.util.stream.Collectors;
  */
 @Component
 @Slf4j
-public class ActionDistributor {
+class ActionDistributor {
 
   // WILL NOT WORK WITHOUT THIS NEXT LINE
   private static final long IMPOSSIBLE_ACTION_ID = 999999999999L;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/DistributionScheduler.java
```diff
@@ -39,6 +39,7 @@ public void run() {
     try {
       distributionInfo = actionDistributorImpl.distribute();
     } catch (Exception e) {
+      distributionInfo = new DistributionInfo();
```
jwgwalton: Is this needed?

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -58,76 +58,71 @@
...
-              actionDistributionListManager.deleteList(actionType.getName(), true);
-            } catch (LockingException e) {
-              log.error("Failed to remove the list of actions just processed from distributed list - "
-                  + "actions distributed OK, but underlying problem may remain");
-              log.error("Stacktrace: ", e);
-            }
-          }
-
-          try {
-            actionDistributionListManager.unlockContainer();
```
jwgwalton: Do we need to still try to unlock the container or will it unlock on it's own? is a try with resources appropriate here?

src/main/java/uk/gov/ons/ctp/response/action/service/impl/ActionPlanJobServiceImpl.java
```diff
@@ -71,84 +68,64 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
     actionPlanRepo.findAll().forEach(actionPlan -> {
-      ActionPlanJob job = new ActionPlanJob();
-      job.setActionPlanFK(actionPlan.getActionPlanPK());
-      job.setCreatedBy(CREATED_BY_SYSTEM);
-      job = createAndExecuteActionPlanJob(job, false);
-      if (job != null) {
-        executedJobs.add(job);
+      Date lastExecutionTime = new Date(nowUTC().getTime() - appConfig.getPlanExecution().getDelayMilliSeconds());
+      if (actionPlan.getLastRunDateTime() == null || actionPlan.getLastRunDateTime().before(lastExecutionTime)) {
+        ActionPlanJob job = ActionPlanJob.builder()
```
jwgwalton: Creating the ActionPlanJob should be in the method createAndExecuteActionPlanJob or that name should be changed to just executeActionPlanJob.

jwgwalton: You could either pass in the ActionPlan or the actionplan pk as the parameter to createAndExecuteActionPlanJob as the only value used off this template is the job.actionPlanFk, which is actually straight off the action plan.

This would make it easier to understand as it's confusing seeing a created actionplanjob passed into a function called createAndExecuteActionPlanJob.

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -58,76 +58,71 @@
...
-                }
-              } catch (Exception e) {
-                log.error("Exception {} thrown processing action {}. Processing will be retried at next scheduled "
-                        + "distribution", e.getMessage(), action.getId());
-                log.error("Stacktrace: ", e);
-              }
-            }
-
-            try {
-              actionDistributionListManager.deleteList(actionType.getName(), true);
```
madeye-matt: Looks to me like the behaviour has been changed here - in the event that any of the action processing throws an exception, the actionDistributionListManager.deleteList call still gets executed.  In the refactored code it won't be called (the deleteList is within the try).  Is this deliberate?

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
benjefferies: `actionDistributionListManager.deleteList(actionType.getName(), true);` will also unlock the container. 

try with resources would be nice but DistributedListManager doesn't implement Closeable but not to worry I am removing DistributedListManager from this logic

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
benjefferies: Good spot. I'll put it in a finally block

src/main/java/uk/gov/ons/ctp/response/action/service/impl/ActionPlanJobServiceImpl.java
benjefferies: To your point I agree accept there is a rest end point which can pass in up to the whole object as a "template"

benjefferies: I was planning to do what you suggested initially until I noticed that

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -58,76 +58,71 @@
...
+
+    try {
+      List<Action> actions = retrieveActions(actionType);
+
+      if (!CollectionUtils.isEmpty(actions)) {
+        processActions(actions, requestCount, cancelCount);
+      }
+    } catch (Exception e) {
+      log.error("Failed to process action type {}", actionType, e);
+      return Arrays.asList(requestCount, cancelCount);
```
madeye-matt: I think this return here is now redundant - if you delete it, the return after the finally will get called and the code will be clearer

benjefferies: Yep you're right, done

PR #30, opened 2018-03-27
Remove redis locking in ActionDistributor

benjefferies:
### Background
The action distributor is responsible for processing actions that have been submitted or cancelled.
A pattern had been implemented to distribute the work amongst multiple instances of the action service by: 
1. Fetching a batch of actions
1. Pushing a set of action IDs to Redis
1. Other services would then check to see i those IDs stored in Redis to determine if they are currently being processed.

### What was changed
This changes that process to remove the dependency on Redis and use  postgres SELECT FOR UPDATE SKIP LOCKED to allow multiple instances of the action service to process actions in parallel

### Why is this changing
The logic used to handle the locking of records via Redis has dependencies on more services, more complicated code which is harder to read and maintain

### How to review and test
The changes shouldn't change any existing functionality so all existing tests and functionality should pass unit and acceptance tests

[trello](https://trello.com/c/qPlEbzaW/68-investigate-use-of-select-for-update-skip-locked-in-place-of-the-current-redis-locking-mechanism-when-using-postgres-as-a-queue)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
@@ -81,10 +81,7 @@ class ActionDistributor {
 
     try {
       final List<Action> actions = retrieveActions(actionType);
-
-      if (!CollectionUtils.isEmpty(actions)) {
-        processActions(actions, requestCount, cancelCount);
-      }
+      processActions(actions, requestCount, cancelCount);
     } catch (final Exception e) {
       log.error("Failed to process action type {}", actionType, e);
     }
@@ -120,10 +117,8 @@ class ActionDistributor {
   private List<Action> retrieveActions(final ActionType actionType) {
     List<Action> actions = actionRepo.findSubmittedOrCancelledByActionTypeName(actionType.getName(),
             appConfig.getActionDistribution().getRetrievalMax());
-    if (CollectionUtils.isNotEmpty(actions)) {
-      log.debug("RETRIEVED action ids {}", actions.stream().map(a -> a.getActionPK().toString())
-          .collect(Collectors.joining(",")));
-    }
+    log.debug("RETRIEVED action ids {}", actions.stream().map(a -> a.getActionPK().toString())
+        .collect(Collectors.joining(",")));
     return actions;
   }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -123,34 +113,17 @@ private void processActions(List<Action> actions, InstructionCount requestCount,
...
-
-    List<Action> actions = actionRepo
-        .findByActionTypeNameAndStateInAndActionPKNotIn(actionType.getName(),
-            Arrays.asList(ActionState.SUBMITTED, ActionState.CANCEL_SUBMITTED), excludedActionIds, pageable);
-    if (!CollectionUtils.isEmpty(actions)) {
+  private List<Action> retrieveActions(ActionType actionType) {
+    List<Action> actions = actionRepo.findSubmittedOrCancelledByActionTypeName(actionType.getName(),
+            appConfig.getActionDistribution().getRetrievalMax());
+    if (CollectionUtils.isNotEmpty(actions)) {
       log.debug("RETRIEVED action ids {}", actions.stream().map(a -> a.getActionPK().toString())
```
jwgwalton: Does this need to be in the empty check, won't it just print "RETRIEVED action ids" if there are no actions.

src/main/resources/application.yml
```diff
@@ -89,7 +89,7 @@ data-grid:
 # the thread that sends actions to handlers
 action-distribution:
   # how many actions should we read each time we wake
-  retrieval-max: 200
+  retrieval-max: 1
```
jwgwalton: ? seems like an accidental change.

benjefferies: Yep was accidentally left there after testing the locking mechanism in postgres. Will revert

JamesGardiner:
@pricem14pc, @davidmort or @zekizeki - this is waiting on approval from one of you and has been sitting open for 44 days. Is it still required/wanted or shall I close it? We can leave the branch there if you think it should be merged in the future.

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -53,12 +42,14 @@
   @Autowired
   private ActionProcessingService actionProcessingService;
 
+
```
mfcrocker: :art:

PR #31, opened 2018-03-28
Fix checkstyle issues

benjefferies:


jwgwalton:
Are we going to adopt this as a platform wide check style?

benjefferies:
@jwgwalton possibly. It's just the checkstyle from https://github.com/ONSdigital/rm-common-config/blob/master/checkstyle/ons-checkstyle-config.xml with a minor tweak for javadoc. The main reason for that is to try and align the codacy errors with our checkstyle errors that we catch when running it locally. I am on to codacy to find out why some things that should be being ignored aren't

PR #39, opened 2018-05-02
add post action rule api

tej99:
Add endpoint to create action rule

Will need: https://github.com/ONSdigital/rm-actionsvc-api/pull/12

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/service/impl/ActionRuleServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/service/impl/ActionRuleServiceImpl.java
@@ -4,6 +4,8 @@ import lombok.extern.slf4j.Slf4j;
 import net.sourceforge.cobertura.CoverageIgnore;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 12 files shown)


tej99:
[TBL123: Scheduling Collection Exercise event dates and Action Service actionplan/job trigger redesign (XXL)](https://trello.com/c/j7zJo9GC/70-tbl123-scheduling-collection-exercise-event-dates-and-action-service-actionplan-job-trigger-redesign-xxl)

src/main/java/uk/gov/ons/ctp/response/action/service/impl/ActionRuleServiceImpl.java
```diff
@@ -19,6 +21,7 @@
 @Slf4j
 public class ActionRuleServiceImpl implements ActionRuleService {
 
+  private static final int TRANSACTION_TIMEOUT = 30;
```
jwgwalton: Shouldn't this be consistent across the service? Could be read in from config.

tej99: It can be but is it worth reading from application.yml file?

PR #14, opened 2018-01-15
Send Notification/Reminder messages to ActionExporter and Notify-gateway

jwgwalton:
- Populates the classifiers and statuses required for sending notifications/generating print letters.
- Validates these messages against the business rules for whether they are to be sent to handler services for letters and emails.

PR #48, opened 2018-06-05
Sts008 receive notification letter

emilioward:
The caseRef was already passed through to the action service this change is setting the caseRef so that it can be used in the pre notification file. 
This change also adds new action types into the action table required for social surveys

**How to test:**

1. Run this along with actionexporter in docker
2. Run `make setup` in the acceptance tests
3. Go into the `action.actionrule` table and you'll need to create a new rule
4. Create a 5th column and add this data to the column - 2, 3, 4, BRICKSOCNOT, Bricks social notification, 10000, 3, b1232e28-7268-481b-b0d0-b4cd8c49e5b1

Like this below:

screen shot 2018-06-07 at 15 31 19

5. Load a sample and CI manually for Bricks 201801 while tailing the logs of the action, actionexporter and collection exercise services. You should ultimately see the CSV being created in the action exporter.

6. Set the collection exercise ready for live

7. Go back to the database in the `action.actionrule` table and set the `daysoffset` back to 0 and save

8. Check the contents of the CSV and see the `caseRef` at the end of the CSV file:

Get the social file name and check it exists with:

`Docker exec sftp ls -lrt /home/centos/Documents/sftp` 

Then plug that social file csv name into this command to check the caseRef is added into the csv:

`Docker exec sftp cat /home/centos/Documents/sftp/SOCIALFILENAMEHERE.csv`

The address columns will be blank since the loaded sample will be a business sample that does not contain address information.

# Links
https://github.com/ONSdigital/rm-actionsvc-api/pull/14
https://github.com/ONSdigital/rm-actionexporter-service/pull/26
https://trello.com/c/xbFrDWWf/100-dev-task-5-int-notify-respondent-of-invitation-to-a-survey-5


emilioward:
[Dev task 5: INT: Notify respondent of invitation to a survey (5)](https://trello.com/c/xbFrDWWf/100-dev-task-5-int-notify-respondent-of-invitation-to-a-survey-5)

src/main/resources/database/changes/release-10.50.0/update_action_plan_add_social.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO action.actiontype
+VALUES
+(4, 'SOCIALNOT', 'Enrolment Invitation Letter', 'Printer', 'n', 'n'),
```
gemmai95: One of us is going to have to change our indexes

tej99: The primary keys will now need changing to 8, 9 and 10.
Can we also add `ON CONFLICT DO NOTHING;` to avoid the application falling over.
And maybe be more explicit here? `INTO action.actiontype(actiontypepk, name, description, handler, cancancel, responserequired)`

benjefferies: I think it should fall over on conflict to indicate the data is being set up wrong?

tej99: Depends if these have been fed in manually in prod or not? Other action plans were already in prod and the scripts are being added in after. I guess these are for social so they shouldn't be in there already...

benjefferies:
@tej99 this should be fixed now

PR #52, opened 2018-06-13
Update search path in schema sql

andrewmil:
# Motivation and Context
Sometimes on startup (this time it was caught in the collection exercise integration tests) the action service will fail to startup because it fails to recognise the `gen_random_uuid()` function in the liquibase scripts. 
# What has changed
This PR sets the search path equal to `action, public` which should give the scripts access to the function

# How to test?
Best to test in association with the integration tests in collection exercise since that's what this has been added in to fix
Should also still pass acceptance tests

# Associated PR's
[rm-collection-exercise](https://github.com/ONSdigital/rm-collection-exercise-service/pull/65)


gemmai95:
[Add creation of action plans to create collection exercise](https://trello.com/c/4rHB8eG3/65-add-creation-of-action-plans-to-create-collection-exercise)

PR #56, opened 2018-06-21
Add action plan selectors

andrewmil:
# Motivation and Context
Action plan selectors are our new solution for determining which action plan a case should be assigned to. These selectors are defined by a general jsonb field which for business surveys are populated by `exerciseRef, surveyRef, activeEnrolment` fields but these could be different for different kinds of surveys.

# What has changed
`actionplanselector` column created with liquibase script
Allow `selectors` field to be passed into POST and PUT requests to `/actionplans(/{id})` which updates the new table
Allow general selectors to be passed in as query string parameters to GET `/actionplans` and retrieve plans based on these selectors
Python script added to the `scripts` folder which creates action plan selectors from existing casetypeoverrides in collection exercise service

# How to test?
`mvn clean install` on the associated [actionsvc-api](https://github.com/ONSdigital/rm-actionsvc-api/tree/add-action-plan-selectors) branch
Try a POST request to `/actionplans` with the selectors field
Try a PUT request to `/actionplans/{id}` with the selectors field for an existing action plan
Try to retrieve an action plan by it's selectors with GET `/actionplans?selector_key=selector_value`
`create-action-plan-selectors` script should create the correct action plan selectors based on the existing casetypeoverrides
Unit/Acceptance tests should pass

# Links
[Trello](https://trello.com/c/piwT43rb/9-create-actionplan-selectors-action)
[actionsvc-api PR](https://github.com/ONSdigital/rm-actionsvc-api/pull/15)
[actionsvc-api PR #2](https://github.com/ONSdigital/rm-actionsvc-api/pull/19)


andrewmil:
@benjefferies It was agreed during the design of this to keep the selectors general because social will require different selectors to business and we weren't 100% on what those were going to be in the future. I think it'd be more complicated if we had two separate paths for social and business, particularly if they can vary between social surveys which I believe they do

src/main/java/uk/gov/ons/ctp/response/action/domain/model/ActionPlanSelector.java
```diff
@@ -0,0 +1,50 @@
...
+      })
+  @Column(name = "actionplanselectorpk")
+  private Integer actionPlanSelectorPK;
+
+  @Column(name = "actionplanfk")
+  private Integer actionPlanFk;
+
+  @Type(type = "jsonb")
+  @Column(name = "selectors", columnDefinition = "jsonb")
+  private HashMap<String, String> selectors;
```
benjefferies: Maybe if this is a one to one mapping it could just be a column on the table?

andrewmil: I was thinking that just when I was finishing this off, I think @pricem14pc's reasoning for it being a separate table was so we wouldn't have that column potentially being null?

andrewmil: Had a chat with Mark, column in the table looks like it makes more sense, I'll look into it now. Should end up being simpler

scripts/create-action-plan-selectors/create-action-plan-selectors.py
```diff
@@ -0,0 +1,77 @@
...
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_casetypeoverrides():
+    logger.debug('Retrieving casetypeoverrides')
+    engine = create_engine(Config.DATABASE_URI)
+    connection = engine.connect()
+
+    sql = "SELECT o.sampleunittypefk, o.actionplanid, c.exerciseref, c.survey_uuid " \
```
shahi645: @andrewmil probably worth adding a comment explaining what the query is doing, also maybe docstrings for the methods in the file

andrewmil: Good point I'll add the comment
Will probably miss docstrings though since this is probably just a one use script

scripts/create-action-plan-selectors/create-action-plan-selectors.py
```diff
@@ -0,0 +1,77 @@
...
+    response.raise_for_status()
+    survey_ref = response.json()['surveyRef']
+    logger.debug('Successfully retrieved surveyRef', survey_id=survey_id, survey_ref=survey_ref)
+    return survey_ref
+
+
+def update_action_plan_selectors(casetypeoverride):
+    action_plan_id = casetypeoverride['actionplanid']
+    exercise_ref = casetypeoverride['exerciseref']
+    sample_unit_type = casetypeoverride["sampleunittypefk"]
```
shahi645: single quotes for consistency, also these 3 things `actionplanid/exerciseref/sampleunittypefk` are always going to be present?

scripts/create-action-plan-selectors/create-action-plan-selectors.py
```diff
@@ -0,0 +1,77 @@
...
+    logger.info('Updating action plan selectors',
+                action_plan_id=action_plan_id, exercise_ref=exercise_ref, sample_unit_type=sample_unit_type)
```
shahi645: Could put these 3 into a variable called `case_attributes`?

andrewmil: They're all only used within this function so don't think there's much benefit in that bringing them into one variable

andrewmil: They all already exist together inside `casetypeoverride`

src/main/java/uk/gov/ons/ctp/response/action/endpoint/ActionPlanEndpoint.java
```diff
@@ -69,20 +105,23 @@
   public final ResponseEntity<ActionPlanDTO> createActionPlan(
```
MIslam92: So theres complaints from the codebeat about the ABC in random points one being here. Would it be worth splitting it up into two separate methods i.e. one that returns the action plan and one that checks if an action plan name already exists?

src/main/java/uk/gov/ons/ctp/response/action/endpoint/ActionPlanEndpoint.java
```diff
@@ -69,20 +105,23 @@
...
       @RequestBody @Valid final ActionPlanPostRequestDTO request, final BindingResult bindingResult)
       throws CTPException, InvalidRequestException {
-    log.info("Create action plan - action plan {}", request);
+    log.info(
+        "Creating action plan, Name: {}, Selectors: {}", request.getName(), request.getSelectors());
     if (bindingResult.hasErrors()) {
       throw new InvalidRequestException("Binding errors for create action plan: ", bindingResult);
     }
 
+    // Check if action plan with same name already exists
```
MIslam92: So theres complaints from the codebeat about the ABC in random points one being here. Would it be worth splitting it up into two separate methods i.e. one that returns the action plan and one that checks if an action plan name already exists?

PR #62, opened 2018-07-23
Removing API.md in favour of swagger docs

benjefferies:
# Motivation and Context
The swagger docs are kept up to date because most of it is generated from the code. It is close to the code and less likely to be forgotten so deprecating the API.md

# What has changed
* Add extra swagger documentation
* Remove API.md

# How to test?
Run the service and navigate to http://localhost:8151/swagger-ui.html
or
`mvn install` and open `target/swagger/paths.md`


benjefferies:
@mfcrocker I guess we don't have to as the documentation has been added to the endpoint classes and will always be accessible here http://rasrm-swagger-dashboard.apps.devtest.onsclofo.uk/

benjefferies:
@mfcrocker I disagree with the statement that using swagger for API documentation makes a repository any less open source but I've added the API.md back in anyway so people have the choice

benjefferies:
@mfcrocker @tej99 files are now checked in

benjefferies:
@damien-lloyd you'll go to https://github.com/ONSdigital/rm-action-service/blob/d666685220833501b050f350c444f71d649a10d6/docs/paths.md (docs/paths.md) instead. I've updated the docs to reflect this

PR #61, opened 2018-07-18
Replace stored procedure

tej99:
# Motivation and Context

Currently a stored procedure is being used to create actions off the back of actionPlanJobs. To give more visibility of what's going on, this function is being shifted from the database layer into the application. 

# What has changed

No more stored procedure. The behaviour is the same just in the JPA repository instead. 

# How to test?

Validate tests pass.

# Links

https://trello.com/c/bj6mBW57/129-remove-stored-procedure-to-create-actions-in-action-service

PR #63, opened 2018-07-24
Revert "Replace stored procedure (#61)"

benjefferies:
# Motivation and Context
We are seeing errors from the insert of the action with `action.gen_random_uuid()` suggesting it doesn't exist. This behaviour doesn't manifest locally and only in CF.

# What has changed
This reverts commit ee4b5a90296722c94c2f32aac8fd310601997c9d. Also, the createactions stored procedure needed to be recreated as it had been dropped

# How to test?
## Test 'concourse-latest' scenario (database has procedure dropped)
- DO NOT recreate docker database (make down)
- SELECT * FROM action.databasechangelog WHERE id = '11-1';
- confirm that line above returns 1 row (if not, go to 'Test production scenario' below)
- checkout and build this branch and deploy to docker
- run social_test_setup
- confirm no errors around 'gen_random_uuid'
## Test 'production' scenario (database procedure not dropped)
- clear out docker environment (make down up)
- run social_test_setup
- confirm no errors around 'gen_random_uuid'


madeye-matt:
That's weird as both worked just fine for me - will pick it up in the morning

tomphp:
FWIW I've just deployed this branch to a space on `devtest` and it started up fine.

PR #66, opened 2018-07-26
Remove gen UUID

tej99:
# Motivation and Context

Currently when we create actions this uses gen_random_uuid, this causes
problems as it's put in different schemas in different environments.

# What has changed

This change switches to an in-memory creation of actions. No more gen_random_uuid.

# How to test?

Run tests

# Links

https://trello.com/c/bj6mBW57/129-remove-stored-procedure-to-create-actions-in-action-service


src/main/java/uk/gov/ons/ctp/response/action/service/impl/ActionServiceImpl.java
```diff
@@ -164,11 +177,40 @@ public Action createAction(final Action action) {
...
   }
 
+  @Transactional
+  @Override
+  public boolean createScheduledActions(Integer actionPlanJobId) {
+    final ActionPlanJob actionPlanJob =
+        actionPlanJobRepository.findByActionPlanJobPK(actionPlanJobId);
+
+    if (actionPlanJob == null) {
+      return true;
```
madeye-matt: Should this be 'false'? Otherwise this method seems to only return true

PurpleBooth: It does only return true, this was the behaviour of the stored proceedure which is why it's still in here. We could remove it though.

PurpleBooth: Removed

PR #67, opened 2018-07-27
Add note about migration naming

PurpleBooth:
Currently the migration names are arbitrary, this adds a structure.


PurpleBooth:
The value is there is no longer misleading version numbers in the migrations that refer to nothing. This makes them explicitly just an incrementing counter.

PR #68, opened 2018-07-27
Cleaned up Maven pom.xml per @atrozzo suggestions

NickGrantONS:
# Motivation and Context
The Maven dependencies were not organised very logically. There were a couple of dependencies which could be removed. Also, the artifactory repos were missing, which meant that developers had to update their local settings.xml to be able to build the project.

# What has changed
1. Added <profiles> section with "artifactory" profile, after the <parent> section
2. Added comment  at top of <dependencies> section
3. Added "spring-boot-starter-web" as first dependency, with tomcat exclusion
4. Grouped all the spring dependencies to the top and added comment  at the bottom of the block
5. Removed "spring-boot-starter-jdbc" dependency
6. Removed "javax.servlet-api" dependency
7. Grouped the ONS libraries together with a comment at the top and the bottom
8. Grouped the 3rd party libraries together with a comment at the top and the bottom
9. Grouped the testing libraries together with a comment at the top and the bottom

# How to test?
Checked file size of the JAR file before & after changes. Ran the acceptance tests - passing.

# Links
Trello: https://trello.com/c/pTzD9hbN/249-refactor-java-service-and-align-the-maven-configuration-removing-redundant-dependenciesm

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,4 +1,6 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -22,159 +24,190 @@
...
-            <groupId>org.springframework.cloud</groupId>
-            <artifactId>spring-cloud-cloudfoundry-connector</artifactId>
+            <groupId>uk.gov.ons.tools</groupId>
+            <artifactId>rabbit</artifactId>
+            <version>1.0.1</version>
         </dependency>
 
+        <!-- ONS END -->
+
+        <!-- third party libraries -->
```
MIslam92: Any reason this can't be Third party libraries to match the closing comment?

NickGrantONS: copy-paste is why  ;o)

PR #72, opened 2018-08-16
Use latest actionsvc-api

tej99:
# Motivation and Context

Latest actionsvc-api uses enum for ActionTypeName, previously it was just a string

# What has changed

Use Enum instead of strings for ActionTypeName

# How to test?

Ran mvn test
and acceptance tests

# Links

requires https://github.com/ONSdigital/rm-actionsvc-api/pull/27

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,7 +150,7 @@
         <dependency>
             <groupId>uk.gov.ons.ctp.product</groupId>
             <artifactId>actionsvc-api</artifactId>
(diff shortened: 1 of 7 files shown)


tej99:
@AdamHawtin @NickGrantONS Should all be passing now, not sure why they were failing before I couldn't replicate it locally.

AdamHawtin:
LGTM, Acceptance tests all passed locally against this branch.

PR #79, opened 2018-08-24
Remove bi/create multiple requests from actions

andrewmil:
# Motivation and Context
When we remove `BI` cases we will only have one `B` case in the action service, event for multiple respondents. We still need to generate the correct actions/notifications from this single case

# What has changed

1. Replaced SQL query for creating actions with code.
2. Made small changes to action distribution process, remove references to parent/child parties and replace with business/respondent parties in relevant places
4. Updated `CaseNotificationService` to handle `ACTIONPLAN_CHANGED` notification types
5. Refactored execution of action plans to require fewer external calls
6. Updated some logging

# How to test?
All tests should pass on `mvn clean install`

Test multiple requests can be created from one action
1. Have two businesses registered for one case (business/survey combination)
2. Case should be on a respondent action plan (formerly BI case action plan)
3. Trigger an action rule for the respondent action plan
4. Two action requests should be created adding messages to the notifygateway queue

Once other BI case work is ready try acceptance tests with all relevant PR's

# Links
[Trello](https://trello.com/c/4Sb2MBAb/302-create-multiple-actions-emails-from-single-b-case-actions-xl)


src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,88 +71,71 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
+  public void createAndExecuteAllActionPlanJobs() {
+    List<ActionPlan> actionPlans = actionPlanRepo.findAll();
+    actionPlans.forEach(
+        actionPlan -> {
+          if (hasActionPlanBeenRunSinceLastSchedule(actionPlan)) {
+            log.with("actionPlanId", actionPlan.getId())
+                .with("actionPlanPK", actionPlan.getActionPlanPK())
+                .debug("Job for plan has been run since last wake up - skipping");
+          } else if (!hasActionableCases(actionPlan)) {
+            log.with("actionPlanId", actionPlan.getId())
```
tej99: We have decided to go with snake case for all keys in logs to match the python services. `action_plan_id`

tej99: Instead of going through and changing all the keys again, you could review https://github.com/ONSdigital/rm-action-service/pull/77 and get that merged in ;)

src/main/java/uk/gov/ons/ctp/response/action/domain/repository/ActionRepository.java
```diff
@@ -97,28 +95,12 @@
...
-          + "   SELECT a "
-          + "   FROM Action a "
-          + "   WHERE a.actionPlanFK = :actionPlanId"
-          + "   AND a.caseId = c.id"
-          + "   AND a.caseFK = c.casePK"
-          + "   AND a.actionRuleFK = r.actionRulePK"
-          + "   AND a.actionType.actionTypePK = r.actionTypeFK )")
-  List<PotentialAction> findPotentialActionsActiveDate(
-      @Param("actionPlanId") Integer actionPlanId, @Param("currentTime") Timestamp currentTime);
+  Action findOneByCaseIdAndActionRuleFKAndPartyId(UUID caseId, Integer actionRuleFK, UUID partyId);
```
NickGrantONS: Soooo much better

src/main/java/uk/gov/ons/ctp/response/action/endpoint/ActionPlanJobEndpoint.java
```diff
@@ -38,14 +34,20 @@
 
   public static final String ACTION_PLAN_JOB_NOT_FOUND = "ActionPlanJob not found for id %s";
 
-  @Autowired private ActionPlanJobService actionPlanJobService;
+  private ActionPlanJobService actionPlanJobService;
+  private ActionPlanService actionPlanService;
```
NickGrantONS: you can make this final

src/main/java/uk/gov/ons/ctp/response/action/endpoint/ActionPlanJobEndpoint.java
```diff
@@ -38,14 +34,20 @@
...
 
-  @Autowired private ActionPlanService actionPlanService;
-
-  @Qualifier("actionBeanMapper")
-  @Autowired
   private MapperFacade mapperFacade;
```
NickGrantONS: make final

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,57 +19,57 @@
...
- * send to downstream handlers. On each wake cycle, it fetches the first n actions of each type, by
- * createddatatime, and forwards them to ActionProcessingService.
- */
+/** This is the service class that distributes actions to downstream services */
 @Component
 class ActionDistributor {
   private static final Logger log = LoggerFactory.getLogger(ActionDistributor.class);
 
-  @Autowired private AppConfig appConfig;
+  private AppConfig appConfig;
```
NickGrantONS: final

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,57 +19,57 @@
...
 
-  @Autowired private ActionRepository actionRepo;
+  private ActionRepository actionRepo;
+  private ActionCaseRepository actionCaseRepo;
+  private ActionTypeRepository actionTypeRepo;
```
NickGrantONS: make final

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,57 +19,57 @@
...
 
-  @Autowired private ActionTypeRepository actionTypeRepo;
-
-  @Autowired
   @Qualifier("business")
   private ActionProcessingService businessActionProcessingService;
```
NickGrantONS: final

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,57 +19,57 @@
...
 
-  @Autowired
   @Qualifier("social")
   private ActionProcessingService socialActionProcessingService;
```
NickGrantONS: final

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -95,26 +92,39 @@ public DistributionInfo distribute() {
...
       try {
         processAction(action, requestCount, cancelCount);
-      } catch (final Exception e) {
-        log.error(
-            "Could not processing action {}."
-                + " Processing will be retried at next scheduled distribution",
-            action.getId(),
-            e);
+      } catch (Exception e) {
+        log.with("actionId", action.getId())
```
NickGrantONS: we're using snake_case for the `.with` keys, to match the Python services

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -95,26 +92,39 @@ public DistributionInfo distribute() {
...
+            .error("Could not process action. Will be retried at next schedule");
```
NickGrantONS: put `, e` on the end to log the exception details

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -95,26 +92,39 @@ public DistributionInfo distribute() {
...
       }
     }
   }
 
+  private void processAction(
+      Action action, final InstructionCount requestCount, final InstructionCount cancelCount)
```
NickGrantONS: don't mix final/non-final unless you're genuinely intending on changing the value of action inside this function (note: final does not mean immutable)

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/DistributionScheduler.java
```diff
@@ -15,25 +12,22 @@
  */
 @CoverageIgnore
 @Component
-public class DistributionScheduler implements HealthIndicator {
+public class DistributionScheduler {
   private static final Logger log = LoggerFactory.getLogger(DistributionScheduler.class);
 
-  private DistributionInfo distributionInfo = new DistributionInfo();
+  private ActionDistributor actionDistributorImpl;
```
NickGrantONS: Rename this - get rid of the Impl

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/DistributionScheduler.java
```diff
@@ -15,25 +12,22 @@
...
 
   /** Scheduled execution of the Action Distributor */
   @Scheduled(fixedDelayString = "#{appConfig.actionDistribution.delayMilliSeconds}")
   public void run() {
-    try {
-      distributionInfo = actionDistributorImpl.distribute();
-    } catch (final Exception e) {
-      log.error("Exception in action distributor", e);
-    }
+    DistributionInfo distInfo = actionDistributorImpl.distribute();
```
NickGrantONS: this is gonna generate a lot of log if it's running every 1 second. Maybe only log if the count is nonzero?

src/main/java/uk/gov/ons/ctp/response/action/scheduled/plan/PlanScheduler.java
```diff
@@ -15,12 +12,14 @@
  * details from the AppConfig
  */
 @Component
-public class PlanScheduler implements HealthIndicator {
+public class PlanScheduler {
   private static final Logger log = LoggerFactory.getLogger(PlanScheduler.class);
 
-  @Autowired private ActionPlanJobService actionPlanJobServiceImpl;
+  private ActionPlanJobService actionPlanJobServiceImpl;
```
NickGrantONS: final

NickGrantONS: also, remove Impl

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -29,17 +27,30 @@
...
-  @Autowired private ActionCaseRepository actionCaseRepo;
-
-  @Autowired private ActionPlanJobRepository actionPlanJobRepo;
-
-  @Autowired private ActionService actionSvc;
+  private AppConfig appConfig;
+
+  private ActionCaseRepository actionCaseRepo;
+  private ActionPlanRepository actionPlanRepo;
+  private ActionPlanJobRepository actionPlanJobRepo;
```
NickGrantONS: make all these member variables final

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,88 +71,71 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
-                    actionPlan.getActionPlanPK());
-              }
-            });
-    return executedJobs;
+  public void createAndExecuteAllActionPlanJobs() {
+    List<ActionPlan> actionPlans = actionPlanRepo.findAll();
+    actionPlans.forEach(
+        actionPlan -> {
+          if (hasActionPlanBeenRunSinceLastSchedule(actionPlan)) {
+            log.with("actionPlanId", actionPlan.getId())
```
NickGrantONS: Use snake_case for keys

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,88 +71,71 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
-                    "Job for plan {} has been run since last wake up - skipping",
-                    actionPlan.getActionPlanPK());
-              }
-            });
-    return executedJobs;
+  public void createAndExecuteAllActionPlanJobs() {
+    List<ActionPlan> actionPlans = actionPlanRepo.findAll();
+    actionPlans.forEach(
+        actionPlan -> {
+          if (hasActionPlanBeenRunSinceLastSchedule(actionPlan)) {
```
NickGrantONS: I would break this if... else out into a separate function, because it's quite a lot of logic to have buried in a lambda

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -40,21 +50,44 @@
...
-  @Autowired private ActionRepository actionRepo;
+  private static final String NO_RESPONDENTS_FOR_SURVEY = "No respondents found for survey";
 
-  @Autowired private ActionCaseRepository actionCaseRepository;
+  private ActionRepository actionRepo;
+  private ActionCaseRepository actionCaseRepo;
+  private ActionPlanRepository actionPlanRepository;
+  private ActionPlanJobRepository actionPlanJobRepository;
+  private ActionRuleRepository actionRuleRepo;
+  private ActionTypeRepository actionTypeRepo;
```
NickGrantONS: Make 'em all final

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -40,21 +50,44 @@
...
 
-  @Autowired private ActionTypeRepository actionTypeRepo;
+  private CollectionExerciseClientService collectionExerciseClientService;
+  private PartySvcClientService partySvcClientService;
```
NickGrantONS: final

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
+          if (isActionPlanLive(caze)) {
+            rules.forEach(
+                rule -> {
+                  if (hasRuleTriggered(rule)) {
+                    try {
+                      createActions(caze, rule);
+                    } catch (Exception ex) {
+                      log.with("caseId", caze.getId().toString())
+                          .with("actionRuleId", rule.getId().toString())
+                          .error(ex.toString());
```
NickGrantONS: Do something like: `.error("Caught exception", ex);` so we get the stack trace etc.

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
+    List<ActionCase> cases = actionCaseRepo.findByActionPlanFK(actionPlan.getActionPlanPK());
+    List<ActionRule> rules = actionRuleRepo.findByActionPlanFK(actionPlan.getActionPlanPK());
+
+    // For each case/rule pair create actions
+    cases.forEach(
+        caze -> {
+          if (isActionPlanLive(caze)) {
+            rules.forEach(
+                rule -> {
+                  if (hasRuleTriggered(rule)) {
```
NickGrantONS: I'd prefer to see the lambda stuff in a function so it can be unit tested more easily, and so it's not so indented

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
+          }
+        });
+    updatePlanAndJob(actionPlan, actionPlanJob);
+  }
 
-    if (actionPlanJob == null) {
-      return;
-    }
+  private boolean isActionPlanLive(ActionCase actionCase) {
+    final Timestamp currentTime = new Timestamp((new Date()).getTime());
```
NickGrantONS: is this time zone safe?

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
+    return triggerDateTime.before(currentTime);
+  }
+
+  private void createActions(ActionCase actionCase, ActionRule actionRule) {
+    ActionType actionType = actionTypeRepo.findByActionTypePK(actionRule.getActionTypeFK());
+
+    // If creating actions for business respondents, create an action per respondent
+    // Else create single action
+    if (actionCase.getSampleUnitType().equals(SampleUnitDTO.SampleUnitType.B.toString())
+        && (actionType.getActionTypePK() == 5 || actionType.getActionTypePK() == 7)) {
```
NickGrantONS: Magic numbers - should be constants

NickGrantONS: In fact, can we look these up in a more sophisticated way, in case the PK changes?

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
 
-    actionRepo
-        .findPotentialActionsActiveDate(actionPlan.getActionPlanPK(), currentTime)
+      List<Association> enrolledAssociations = respondentsEnrolledOnCase(actionCase);
+      if (enrolledAssociations.isEmpty()) {
```
NickGrantONS: Can it ever be null too?

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
+  }
+
+  private boolean isAssociationEnabledForSurvey(Association association, String surveyId) {
+    return association
+        .getEnrolments()
+        .stream()
+        .anyMatch(
+            enrolment ->
+                enrolment.getSurveyId().equals(surveyId)
+                    && enrolment.getEnrolmentStatus().equalsIgnoreCase(ENABLED));
```
NickGrantONS: Where is ENABLED actually defined - it must be in an Enum somewhere, isn't it?

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
+  }
+
+  private void createAction(
+      ActionCase actionCase, ActionRule actionRule, ActionType actionType, UUID partyId) {
+    if (actionRepo.findOneByCaseIdAndActionRuleFKAndPartyId(
```
NickGrantONS: if this is the only place the repo method is used, you should change it to `exists`

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
+          .debug("Action already exists");
+      return;
+    }
+
+    Action newAction = new Action();
+    newAction.setId(UUID.randomUUID());
+    newAction.setCreatedBy("SYSTEM");
+    newAction.setManuallyCreated(false);
+    newAction.setState(ActionState.SUBMITTED);
+    newAction.setCreatedDateTime(new Timestamp((new Date()).getTime()));
```
NickGrantONS: is this timezone safe? You used a function called `nowUTC()` somewhere else in the code

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
+      log.with("caseId", actionCase.getId().toString())
+          .with("actionRuleId", actionRule.getId().toString())
+          .with("partyId", partyId.toString())
+          .debug("Action already exists");
+      return;
+    }
+
+    Action newAction = new Action();
+    newAction.setId(UUID.randomUUID());
+    newAction.setCreatedBy("SYSTEM");
```
NickGrantONS: Use the force, Luke. Also, string constant

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -173,25 +200,129 @@ public Action createAction(final Action action) {
...
+    newAction.setActionType(actionType);
+    newAction.setPriority(actionRule.getPriority());
 
+    newAction.setPartyId(partyId);
+
+    actionRepo.saveAndFlush(newAction);
+  }
+
+  private void updatePlanAndJob(ActionPlan actionPlan, ActionPlanJob actionPlanJob) {
+    final Timestamp currentTime = new Timestamp((new Date()).getTime());
```
NickGrantONS: Time zone safe?

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -18,108 +17,134 @@
...
+ * action.case table
  */
 @Service
 public class CaseNotificationService {
   private static final Logger log = LoggerFactory.getLogger(CaseNotificationService.class);
 
   private static final int TRANSACTION_TIMEOUT = 30;
 
-  @Autowired private ActionCaseRepository actionCaseRepo;
+  private ActionCaseRepository actionCaseRepo;
```
NickGrantONS: final

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -18,108 +17,134 @@
...
 
-  @Autowired private ActionPlanRepository actionPlanRepo;
+  private ActionPlanRepository actionPlanRepo;
 
-  @Autowired private ActionService actionService;
+  private ActionService actionService;
 
-  @Autowired private CollectionExerciseClientService collectionSvcClientServiceImpl;
+  private CollectionExerciseClientService collectionSvcClientServiceImpl;
```
NickGrantONS: final and remove Impl

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -18,108 +17,134 @@
...
+        break;
+
+      case DISABLED:
+      case DEACTIVATED:
+        actionService.cancelActions(caseId);
+        deleteActionCase(caseId);
+        break;
+
+      default:
+        log.with("notification_type", notification.getNotificationType())
```
NickGrantONS: This should probably be a `throw new IllegalArgumentException("Invalid notification type")`

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -18,108 +17,134 @@
...
-    } else {
-      log.warn("Cannot accept CaseNotification for non-existent actionplan {}", actionPlanIdStr);
+    actionCaseRepo.flush();
+  }
+
+  private ActionCase createActionCase(CaseNotification notification) throws CTPException {
+    UUID actionPlanId = UUID.fromString(notification.getActionPlanId());
+    ActionPlan actionPlan = actionPlanRepo.findById(actionPlanId);
+    if (actionPlan == null) {
+      throw new CTPException(CTPException.Fault.RESOURCE_NOT_FOUND, actionPlanId.toString());
```
NickGrantONS: 404 when trying to POST would be a bit weird. What's the logic here? Might be more like an `IllegalStateException`

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -18,108 +17,134 @@
...
+    UUID collectionExerciseId = UUID.fromString(notification.getExerciseId());
+    UUID partyId =
+        notification.getPartyId() == null ? null : UUID.fromString(notification.getPartyId());
+    UUID sampleUnitId =
+        notification.getSampleUnitId() == null
+            ? null
+            : UUID.fromString(notification.getSampleUnitId());
 
-    actionCaseRepo.flush();
+    final CollectionExerciseDTO collectionExercise =
```
NickGrantONS: one lone final looks out of place

src/main/java/uk/gov/ons/ctp/response/action/service/decorator/context/BusinessActionRequestContextFactory.java
```diff
@@ -22,88 +20,70 @@
   private static final Logger log =
       LoggerFactory.getLogger(BusinessActionRequestContextFactory.class);
 
-  @Autowired private PartySvcClientService partySvcClientService;
+  private final PartySvcClientService partySvcClientService;
 
-  @Autowired private DefaultActionRequestContextFactory defaultFactory;
+  private final DefaultActionRequestContextFactory defaultFactory;
```
NickGrantONS: final

src/main/java/uk/gov/ons/ctp/response/action/service/decorator/context/BusinessActionRequestContextFactory.java
```diff
@@ -22,88 +20,70 @@
...
       setPartiesForBusiness(context);
     }
 
     return context;
   }
 
-  private boolean isIndividual(ActionRequestContext context) {
-    return context.getSampleUnitType().isParent() == false;
+  private boolean isRespondent(ActionRequestContext context) {
+    return context.getAction().getActionRuleFK() == 5 || context.getAction().getActionRuleFK() == 7;
```
NickGrantONS: Magic numbers.

src/main/java/uk/gov/ons/ctp/response/action/service/decorator/context/BusinessActionRequestContextFactory.java
```diff
@@ -22,88 +20,70 @@
...
   }
 
-  public List<PartyDTO> getChildParties(
-      final PartyDTO parentParty, final SampleUnitType parentUnitType) {
-    List<PartyDTO> childParties = new ArrayList<>();
-
-    final List<String> childPartyIds =
-        parentParty
+  public List<PartyDTO> getRespondentParties(final PartyDTO businessParty) {
+    final List<String> respondentPartyIds =
```
NickGrantONS: you're using a mix of final/non-final. Need to be consistent

src/main/java/uk/gov/ons/ctp/response/action/service/decorator/context/DefaultActionRequestContextFactory.java
```diff
@@ -22,13 +21,22 @@
...
-  @Autowired private CaseSvcClientService caseSvcClientService;
-
-  @Autowired private CollectionExerciseClientService collectionExerciseClientService;
-
-  @Autowired private SurveySvcClientService surveySvcClientService;
+  private ActionPlanRepository actionPlanRepo;
+
+  private CaseSvcClientService caseSvcClientService;
+  private CollectionExerciseClientService collectionExerciseClientService;
+  private SurveySvcClientService surveySvcClientService;
```
NickGrantONS: final

src/main/resources/database/changelog-master.yml
```diff
@@ -46,3 +46,6 @@ databaseChangeLog:
 
   - include:
         file: database/changes/release-11/changelog.yml
+
+  - include:
+      file: database/changes/release-12/changelog.yml
```
NickGrantONS: looks like the indenting is different

src/test/java/uk/gov/ons/ctp/response/action/endpoint/ActionPlanJobEndpointUnitTest.java
```diff
@@ -290,76 +283,4 @@ public void findActionPlanJob() throws Exception {
...
-        .andExpect(jsonPath("$.actionPlanId", is(ACTIONPLANID.toString())))
-        .andExpect(jsonPath("$.createdBy", is(CREATED_BY_SYSTEM)))
-        .andExpect(jsonPath("$.state", is(ActionPlanJobDTO.ActionPlanJobState.COMPLETED.name())))
-        .andExpect(
-            jsonPath(
-                "$.createdDateTime", is(new DateMatcher(CREATED_DATE_TIME_ACTION_PLAN_JOB_ID_1))))
-        .andExpect(
-            jsonPath(
-                "$.updatedDateTime", is(new DateMatcher(UPDATED_DATE_TIME_ACTION_PLAN_JOB_ID_1))));
-  }
```
NickGrantONS: These look like valid tests to me

src/test/java/uk/gov/ons/ctp/response/action/endpoint/ActionPlanJobEndpointUnitTest.java
```diff
@@ -290,76 +283,4 @@ public void findActionPlanJob() throws Exception {
...
-                ACTION_PLAN_JOB_JSON_FAILING_VALIDATION));
-
-    actions
-        .andExpect(status().isBadRequest())
-        .andExpect(handler().handlerType(ActionPlanJobEndpoint.class))
-        .andExpect(handler().methodName("executeActionPlan"))
-        .andExpect(jsonPath("$.error.code", is(CTPException.Fault.VALIDATION_FAILED.name())))
-        .andExpect(jsonPath("$.error.message", is(INVALID_JSON)))
-        .andExpect(jsonPath("$.error.timestamp", isA(String.class)));
-  }
```
NickGrantONS: Isn't this a potentially valid test?

src/test/java/uk/gov/ons/ctp/response/action/service/ActionProcessingServiceTest.java
```diff
@@ -217,170 +304,35 @@ public void testProcessActionRequestActionPersistingActionThrowsException() thro
...
-            eq(ACTIONEXPORTER),
+            any(String.class),
             any(uk.gov.ons.ctp.response.action.message.instruction.Action.class));
-  }
-
-  /**
-   * Happy path for an action linked to a case for a CHILD sample unit (a HI one), ie we go all the
-   * way to producing an ActionRequest and publishing it.
-   */
-  @Test
```
NickGrantONS: What's "HI" - is that social individual or something? Do we need this test?

src/test/java/uk/gov/ons/ctp/response/action/service/ActionServiceTest.java
```diff
@@ -194,80 +212,76 @@ public void testUpdateActionNoUpdate() {
...
+    verify(actionRepo, times(1)).saveAndFlush(any());
+    verify(actionPlanRepo, times(1)).saveAndFlush(any());
+    verify(actionPlanJobRepo, times(1)).saveAndFlush(any());
+  }
 
-    // did not attempt to create actions
-    verify(actionRepo, times(0)).save(any(Action.class));
-    verify(actionRepo, times(1)).flush();
+  @Test
+  public void testCreateScheduledActionsBCaseRule5() {
```
NickGrantONS: Does "Rule 5" mean PK = 5? We probably shouldn't hard-code primary keys into our test names etc.

src/test/java/uk/gov/ons/ctp/response/action/service/ActionServiceTest.java
```diff
@@ -194,80 +212,76 @@ public void testUpdateActionNoUpdate() {
...
-    // action plan has been updated
-    verify(actionPlan, times(1)).setLastRunDateTime(any(Timestamp.class));
-
-    // action plan job has been updated
-    verify(actionPlanJob, times(1)).complete(any(Timestamp.class));
-
-    // attempted to create actions
-    verify(actionRepo, times(1)).save(any(Action.class));
-    verify(actionRepo, times(1)).flush();
+  public void testCreateScheduledActionsBCaseRule5NoEnrolledRespondents() {
```
NickGrantONS: gotta be a better name than "rule 5"

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
andrewmil: So in the python party service I think it is defined as an Enum but I can't see that anywhere in the java implementation.
There's a `rm-party-service-api` where it's just defined as a String.

src/test/java/uk/gov/ons/ctp/response/action/service/ActionProcessingServiceTest.java
andrewmil: Talked to Dave about this from capabilities. HI cases aren't in use yet and we're good to remove the depreciated functionality for it

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
andrewmil: Making this final was having strange effects on the tests. It seemed a bit challenging to fix the tests around it so leaving the `final` out for now

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
andrewmil: Making these final was having strange effects on the tests. It seemed a bit challenging to fix the tests around it so leaving the `final` out for now

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
andrewmil: :+1: You're right, here we should be checking on the Type Handler rather than typePK

src/main/java/uk/gov/ons/ctp/response/action/service/decorator/PartyAndContact.java
```diff
@@ -62,7 +59,12 @@ public void decorateActionRequest(ActionRequest actionRequest, ActionRequestCont
     }
 
     actionRequest.setContact(actionContact);
-    actionRequest.setEnrolmentStatus(getEnrolmentStatus(parentParty));
+    actionRequest.setEnrolmentStatus(getEnrolmentStatus(businessParty));
+  }
+
+  private boolean isRespondent(ActionRequestContext context) {
+    return context.getAction().getActionType().getActionTypePK() == 5
+        || context.getAction().getActionType().getActionTypePK() == 7;
```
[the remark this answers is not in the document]
andrewmil: 👍 Now checking on the ActionType handler rather than TypePK

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
andrewmil: Making these final was having strange effects on the tests. It seemed a bit challenging to fix the tests around it so leaving the final out for now

src/main/java/uk/gov/ons/ctp/response/action/domain/repository/ActionRepository.java
```diff
@@ -102,5 +102,5 @@
    * @param partyId UUID of associated party
    * @return Action returns an action is
    */
-  Action findOneByCaseIdAndActionRuleFKAndPartyId(UUID caseId, Integer actionRuleFK, UUID partyId);
+  Boolean existsByCaseIdAndActionRuleFKAndPartyId(UUID caseId, Integer actionRuleFK, UUID partyId);
```
NickGrantONS: you can use the primitive boolean, and you should prefer it because it only has two possible values (true and false) instead of three (Boolean.TRUE, Boolean.FALSE and null)

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -18,108 +17,134 @@
...
+
+  private void updateActionCase(UUID actionCaseId, UUID actionPlanId) throws CTPException {
+    ActionCase existingCase = actionCaseRepo.findById(actionCaseId);
+    if (existingCase == null) {
+      log.with("case_id", actionCaseId.toString()).error("No case found to update");
+      throw new CTPException(CTPException.Fault.RESOURCE_NOT_FOUND);
+    }
+    ActionPlan actionPlan = actionPlanRepo.findById(actionPlanId);
+    if (actionPlan == null) {
+      throw new CTPException(CTPException.Fault.RESOURCE_NOT_FOUND, actionPlanId.toString());
```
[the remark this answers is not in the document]
andrewmil: Since this code is called from a queue rather than an API endpoint, all of these CTPException's are being replaced with IllegalStateException's

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
andrewmil: I think this should default to UTC which is what we want to be saving in the database?

src/test/java/uk/gov/ons/ctp/response/action/endpoint/ActionPlanJobEndpointUnitTest.java
andrewmil: This endpoint has been updated to not accept JSON input anymore. 
It wasn't being used for anything I could see

src/test/java/uk/gov/ons/ctp/response/action/endpoint/ActionPlanJobEndpointUnitTest.java
andrewmil: No longer accepting any JSON input for this endpoint

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
NickGrantONS: Still... DateTimeUtil.nowUTC() is used elsewhere to get a timestamp. Make it consistent?

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -137,11 +144,11 @@ private void updateActionCase(UUID actionCaseId, UUID actionPlanId) throws CTPEx
     actionCaseRepo.save(existingCase);
   }
 
-  private void deleteActionCase(UUID actionCaseId) throws CTPException {
+  private void deleteActionCase(UUID actionCaseId) {
     ActionCase actionCaseToDelete = actionCaseRepo.findById(actionCaseId);
     if (actionCaseToDelete == null) {
       log.with("case_id", actionCaseId.toString()).error("No case found to delete");
```
NickGrantONS: maybe change this to `.warn`

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -176,25 +205,129 @@ public Action createAction(final Action action) {
...
+    newAction.setCaseId(actionCase.getId());
+
+    newAction.setActionPlanFK(actionRule.getActionPlanFK());
+    newAction.setActionRuleFK(actionRule.getActionRulePK());
+    newAction.setActionType(actionType);
+    newAction.setPriority(actionRule.getPriority());
+
+    newAction.setPartyId(partyId);
+
+    actionRepo.saveAndFlush(newAction);
```
tej99: Is it sensible to have an info level log here? I know it's a happy path but would be very useful when debugging this complex process of creating actions. The alternative would be to go into the prod database and query the actions table to see if an action has been created (not everyone has this access).

If not info there should at least be a debug log line for the successful action creation.

NickGrantONS: debug not info

src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
```diff
@@ -53,14 +57,31 @@ public ActionProcessingService(ActionRequestDecorator[] decorators) {
...
 
-  public ActionRequest prepareActionRequest(Action action) {
+  public List<ActionRequest> prepareActionRequests(Action action) {
     final ActionRequestContextFactory factory = getActionRequestDecoratorContextFactory();
     final ActionRequestContext context = factory.getActionRequestDecoratorContext(action);
-    final ActionRequest actionRequest = new ActionRequest();
 
-    Arrays.stream(this.decorators).forEach(d -> d.decorateActionRequest(actionRequest, context));
+    ArrayList<ActionRequest> actionRequests = new ArrayList<>();
+    if (context.getCaseDetails().getSampleUnitType().equals("B")
```
NickGrantONS: Use the enum instead of hard-coded string?

src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
```diff
@@ -53,14 +57,31 @@ public ActionProcessingService(ActionRequestDecorator[] decorators) {
...
+        && context.getAction().getActionType().getHandler().equals("Notify")) {
```
NickGrantONS: hard coded string

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -226,14 +212,14 @@ private boolean isActionPlanLive(ActionCase actionCase) {
...
+  private void createActionForCaseAndRule(ActionCase actionCase, ActionRule actionRule) {
     if (hasRuleTriggered(actionRule)) {
       try {
-        createActions(actionCase, actionRule);
+        createAction(actionCase, actionRule);
       } catch (Exception ex) {
         log.with("case_id", actionCase.getId().toString())
             .with("action_rule_id", actionRule.getId().toString())
-            .error("Failed to create actions", ex);
+            .error("Failed to create action", ex);
```
NickGrantONS: I'm a bit uncomfortable about logging errors and not rethrowing them - what if the exception was due to a temporary blip contacting Postgres? We'd want the whole transaction rolled back so that we could retry, wouldn't we?

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -86,29 +81,47 @@ public DistributionInfo distribute() {
             .build();
 
     try {
-      final List<Action> actions = retrieveActions(actionType);
+      List<Action> actions = retrieveActions(actionType);
       processActions(actions, requestCount, cancelCount);
-    } catch (final Exception e) {
-      log.with("action_type", actionType.getName()).error("Failed to process action type", e);
+    } catch (Exception e) {
+      log.with("action_type", actionType).error("Failed to process", e);
```
NickGrantONS: I'm worried about this... catching all exceptions like this means we never rollback, which means we might not retry properly in the event of a temporary glitch - we're not letting things unwind themselves elegantly

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -176,25 +191,76 @@ public Action createAction(final Action action) {
...
+    final Timestamp currentTime = new Timestamp((new Date()).getTime());
+    return actionCase.getActionPlanStartDate().before(currentTime)
+        && actionCase.getActionPlanEndDate().after(currentTime);
+  }
+
+  private void createActionForCaseAndRule(ActionCase actionCase, ActionRule actionRule) {
+    if (hasRuleTriggered(actionRule)) {
+      try {
+        createAction(actionCase, actionRule);
+      } catch (Exception ex) {
```
NickGrantONS: I'm always pretty afraid of any catch-all exception handler. Any reason why we have this here?

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -18,109 +17,141 @@
...
+      case DISABLED:
+      case DEACTIVATED:
+        actionService.cancelActions(caseId);
+        deleteActionCase(caseId);
+        break;
+
+      default:
+        log.with("notification_type", notification.getNotificationType())
+            .warn("Unknown case notification type", notification.getNotificationType());
+        throw new IllegalArgumentException("Invalid notification type");
```
NickGrantONS: I would log this as `.error` because it will actually block the Rabbit queue until the bad message is DLQ'ed

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,87 +71,76 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
+          }
+        });
+  }
+
+  private boolean shouldCreateAndExecuteActionPlanJob(ActionPlan actionPlan) {
+    if (hasActionPlanBeenRunSinceLastSchedule(actionPlan)) {
+      log.with("action_plan_id", actionPlan.getId())
+          .debug("Job for plan has been run since last wake up - skipping");
+      return false;
+    } else if (!hasActionableCases(actionPlan)) {
```
atrozzo: a bit of overkill all these conditions for a debug level comment ( ? )

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,87 +71,76 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
+    actionPlans.forEach(
+        actionPlan -> {
+          if (shouldCreateAndExecuteActionPlanJob(actionPlan)) {
+            createAndExecuteActionPlanJob(actionPlan);
+          }
+        });
+  }
+
+  private boolean shouldCreateAndExecuteActionPlanJob(ActionPlan actionPlan) {
+    if (hasActionPlanBeenRunSinceLastSchedule(actionPlan)) {
```
atrozzo: the method stink a bit, considering is checking conditions and does not change any status. The multiple if for logging debug informations are quite bad. What about check hasActionableCases || hasActionPlanBeenRunSinceLastSchedule as false condition otherwise return true ?

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,87 +71,76 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
-      log.with("action_plan_pk", actionPlanPK).debug("Action plan is null");
-      return null;
-    }
-
-    if (actionCaseRepo.countByActionPlanFK(actionPlanPK) == 0) {
-      log.with("action_plan_pk", actionPlanPK).debug("No open cases for action plan");
-
-      return null;
-    }
+  public ActionPlanJob createAndExecuteActionPlanJob(final ActionPlan actionPlan) {
```
atrozzo: SRP in name convention create and execute.

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -51,74 +56,37 @@ public ActionDistributor(
   /**
    * Called on schedule to check for submitted actions then creates and distributes the actions to
    * action exporter or notify gateway
-   *
-   * @return the info for the health endpoint regarding the distribution just performed
    */
-  @Transactional
-  public DistributionInfo distribute() {
-    final DistributionInfo distInfo = new DistributionInfo();
+  public void distribute() {
```
NickGrantONS: I wonder if this will have problems because there's no transaction - even read-only - which wraps the findall of the action types etc.

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -51,74 +56,37 @@ public ActionDistributor(
...
-    final InstructionCount cancelCount =
-        InstructionCount.builder()
-            .actionTypeName(actionType.getName())
-            .instruction(DistributionInfo.Instruction.CANCEL_REQUEST)
-            .count(0)
-            .build();
-
+  private void processActionType(final ActionType actionType) {
+    String actionTypeName = actionType.getName();
+    RLock lock = redissonClient.getFairLock(lockPrefix + actionTypeName);
```
NickGrantONS: you need to lock the lock after you get it

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -111,6 +100,7 @@ private boolean hasActionPlanBeenRunSinceLastSchedule(ActionPlan actionPlan) {
    * @param actionPlan Action plan to create and execute job for
    * @return ActionPlanJob that was created or null if it has not been created.
    */
+  @Transactional
```
NickGrantONS: This annotation won't actually do anything because the method - technically - isn't 'public' per se... it could be made private. What this means is that Spring won't be able to intercept the method call. You'll have to put `@Transactional` on one of the 'entry points' into this bean

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
andrewmil: 🤦‍♂️ yup that would make sense

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -66,12 +69,24 @@ private void processActionType(final ActionType actionType) {
...
-      lock.unlock();
+      // Wait for a lock. Automatically unlock after a certain amount of time to prevent issues
+      // when lock holder crashes or Redis crashes causing permanent lockout
+      if (lock.tryLock(
+          appConfig.getDataGrid().getLockTimeToWaitSeconds(),
+          appConfig.getDataGrid().getLockTimeToLiveSeconds(),
+          TimeUnit.SECONDS)) {
+        try {
+          List<Action> actions =
+              actionRepo.findSubmittedOrCancelledByActionTypeName(
```
NickGrantONS: This could actually be improved a lot. At the moment it only grabs a small batch (`retrievalMax`) and it could actually get a stream to _all_ the actions in the states we need - we'd need a new repo method, but it means we could get rid of the native Postgres query and the DB locking

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -72,6 +72,7 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
     return actionPlanJobRepo.findByActionPlanFK(actionPlan.getActionPlanPK());
   }
 
+  @Transactional
```
NickGrantONS: This needs `propagation = REQUIRES_NEW`

src/main/resources/application-cloud.yml
```diff
@@ -83,6 +83,8 @@ data-grid:
   list-time-to-wait-seconds: 600
   #  after app death how long should the lock on all lists remain
   lock-time-to-live-seconds: 600
+  #  how long to wait to get the lock
+  lock-time-to-wait-seconds: 600
```
NickGrantONS: time to live probably needs to be a lot more

src/main/resources/application-test.yml
```diff
@@ -25,6 +25,8 @@ data-grid:
   list-time-to-wait-seconds: 10
   #  after app death how long should the lock on all lists remain
   lock-time-to-live-seconds: 10
+  #  how long to wait to get the lock
+  lock-time-to-wait-seconds: 10
```
NickGrantONS: you actually only need this if you're overriding values in `application.yml`

src/test/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributorTest.java
```diff
@@ -85,10 +80,18 @@ public void setUp() throws Exception {
     actionCase = new ActionCase();
     actionCase.setSampleUnitType("H");
 
+    MockitoAnnotations.initMocks(this);
+    DataGrid dataGrid = new DataGrid();
+    dataGrid.setLockTimeToLiveSeconds(30);
+    dataGrid.setLockTimeToWaitSeconds(600);
+    when(appConfig.getDataGrid()).thenReturn(dataGrid);
```
NickGrantONS: alternatively, you could've returned a mock `DataGrid` and intercepted the `getLockTime` methods.

src/test/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributorTest.java
```diff
@@ -85,10 +80,18 @@ public void setUp() throws Exception {
...
+    ActionDistribution actionDistribution = new ActionDistribution();
+    actionDistribution.setRetrievalMax(1000);
+    when(appConfig.getActionDistribution()).thenReturn(actionDistribution);
+
     lock = mock(RLock.class);
     when(redissonClient.getFairLock(any())).thenReturn(lock);
-
-    MockitoAnnotations.initMocks(this);
+    when(lock.tryLock(anyLong(), anyLong(), any(TimeUnit.class))).thenReturn(true);
```
NickGrantONS: you probably want to `.verify` that `.tryLock` and `.unlock` are actually called

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -98,21 +99,18 @@ private boolean hasActionPlanBeenRunSinceLastSchedule(ActionPlan actionPlan) {
    * Create a new ActionPlanJob and execute associated actions
    *
    * @param actionPlan Action plan to create and execute job for
-   * @return ActionPlanJob that was created or null if it has not been created.
    */
-  @Transactional
-  public ActionPlanJob createAndExecuteActionPlanJob(final ActionPlan actionPlan) {
+  private void createAndExecuteActionPlanJob(final ActionPlan actionPlan) {
 
     if (!actionPlanExecutionLockManager.lock(actionPlan.getName())) {
```
NickGrantONS: I don't think you need this lock anymore, because there's a lock in the scheduler which will protect any contention from other processes

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
andrewmil: :+1: Will look into this after prunes test

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -88,20 +86,8 @@ private boolean hasActionableCases(ActionPlan actionPlan) {
...
-  }
-
-  /**
-   * Create a new ActionPlanJob and execute associated actions
-   *
-   * @param actionPlan Action plan to create and execute job for
-   */
-  private void createAndExecuteActionPlanJob(final ActionPlan actionPlan) {
-
+  @Transactional(propagation = REQUIRES_NEW)
```
NickGrantONS: this transaction hint will be ignored when this method gets called internally (and it's more like a private method anyway)

src/test/java/uk/gov/ons/ctp/response/action/service/decorator/PartyActionRequestDecoratorTest.java
```diff
@@ -1,131 +1,131 @@
...
-    businessAttributes.setTradstyle3("TRADSTYLE3");
-
-    final String generatedTradingStyle =
-        this.partyActionRequestDecorator.generateTradingStyle(businessAttributes);
-    final String expectedTradingStyle = "TRADSTYLE1 TRADSTYLE3";
-
-    assertEquals(expectedTradingStyle, generatedTradingStyle);
-  }
-}
+// package uk.gov.ons.ctp.response.action.service.decorator;
```
NickGrantONS: I presume you're planning on getting rid of this class before you merge it?

andrewmil: Working on fixing these tests today, will get them up tomorrow

src/main/java/uk/gov/ons/ctp/response/action/service/decorator/PartyAndContact.java
```diff
@@ -15,88 +15,101 @@
...
-    actionRequest.setRespondentStatus(parseRespondentStatuses(context.getChildParties()));
-    List<PartyDTO> createdChildParties =
-        filterListByStatus(context.getChildParties(), ActionProcessingService.CREATED);
+    if (enrolmentStatuses.contains(ActionProcessingService.PENDING)) {
+      enrolmentStatus = ActionProcessingService.PENDING;
+    }
 
-    if (createdChildParties != null && createdChildParties.size() > 0) {
-      actionRequest.setIac(""); // Don't want to send this to the business, breaks if null
+    if (enrolmentStatuses.contains(ActionProcessingService.ENABLED)) {
```
NickGrantONS: it would be more efficient to do this test first, then put the second test in an `else if` so that it doesn't waste time doing two string comparisons and two variable assignments, when it could only do one of each.

scripts/migration-replace-bi-cases-with-b/temp_cases.csv
```diff
@@ -0,0 +1,33 @@
+7bd9c119-fa4c-41d9-a317-87d8034681d7,d30751e0-492a-413f-8879-9735fcafc014
```
NickGrantONS: this shouldn't be in Git

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,101 +22,105 @@
...
- * createddatatime, and forwards them to ActionProcessingService.
- */
+/** This is the service class that distributes actions to downstream services */
 @Component
 class ActionDistributor {
 
   private static final Logger log = LoggerFactory.getLogger(ActionDistributor.class);
 
-  @Autowired private AppConfig appConfig;
+  private static final String lockPrefix = "ActionDistributionLock-";
```
NickGrantONS: constants should be CAPITAL_CASE

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,101 +22,105 @@
...
 
-  @Autowired private ActionRepository actionRepo;
+  private AppConfig appConfig;
```
NickGrantONS: you can't make this final?

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,101 +22,105 @@
...
+  private RedissonClient redissonClient;
 
-  @Autowired private ActionTypeRepository actionTypeRepo;
+  private ActionRepository actionRepo;
```
NickGrantONS: make final?

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,101 +22,105 @@
...
+    RLock lock = redissonClient.getFairLock(lockPrefix + actionTypeName);
     try {
-      final List<Action> actions = retrieveActions(actionType);
-      processActions(actions, requestCount, cancelCount);
-    } catch (final Exception e) {
-      log.with("action_type", actionType.getName()).error("Failed to process action type", e);
+      // Wait for a lock. Automatically unlock after a certain amount of time to prevent issues
+      // when lock holder crashes or Redis crashes causing permanent lockout
+      if (lock.tryLock(
+          appConfig.getDataGrid().getLockTimeToWaitSeconds(),
```
NickGrantONS: in actual fact - and I got this wrong in Sample - it might be preferable to skip the wait, because if one type is locked then we can just work on another one, without waiting for the other process to relinquish the lock. It'll be more performant

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
NickGrantONS: you gonna have a look at refactoring it now?

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,101 +22,105 @@
...
+      caseSvcClientService.generateNewIacForCase(action.getCaseId());
+    }
+
+    try {
+      if (action.getState().equals(ActionState.SUBMITTED)) {
+        ap.processActionRequests(action);
+      } else if (action.getState().equals(ActionState.CANCEL_SUBMITTED)) {
+        ap.processActionCancel(action);
       }
+    } catch (CTPException ex) {
```
NickGrantONS: I hate CTPException so much. I actually catch it at the point where it's thrown and re-throw it as something useful like IllegalStateException, so we don't have to pollute our code with loads of useless `throws` declarations and `try...catch`es.

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,87 +67,54 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
+          if (hasActionableCases(actionPlan)) {
+            createAndExecuteActionPlanJob(actionPlan);
+          }
+        });
+  }
 
-      return null;
-    }
+  private boolean hasActionableCases(ActionPlan actionPlan) {
+    return actionCaseRepo.countByActionPlanFK(actionPlan.getActionPlanPK()) > 0;
```
NickGrantONS: if this is the only place this repo method is used, consider refactoring it to be `exists` instead of `count`

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,87 +67,54 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
+  }
 
+  @Transactional(propagation = REQUIRES_NEW)
+  public void createAndExecuteActionPlanJob(final ActionPlan actionPlan) {
     if (!actionPlanExecutionLockManager.lock(actionPlan.getName())) {
-      log.with("action_plan_pk", actionPlanPK).debug("Could not get lock on action plan");
-
-      return null;
+      log.with("action_plan_id", actionPlan.getId()).debug("Could not get manager lock");
```
NickGrantONS: so... this will *definitely* be retried, right? We don't want to carry on thinking this has been done OK if it hasn't

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanService.java
```diff
@@ -68,7 +68,9 @@ public ActionPlan createActionPlan(final ActionPlan actionPlan) {
         .with("action_plan_name", actionPlan.getName())
         .debug("Creating action plan");
 
-    ActionPlan savedActionPlan = saveActionPlan(actionPlan);
+    actionPlan.setActionPlanPK(null);
```
NickGrantONS: a primary key of null feels very wrong

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
andrewmil: Makes sense I'll remove the wait

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
andrewmil: If we fail to get the lock here it will retry at next schedule

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanService.java
andrewmil: `actionPlanPK` is set from the sequence in the database.
I think this `null` is set to make sure that we're not using a PK sent in the request

src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
```diff
@@ -20,20 +22,20 @@
...
   public static final String CANCELLATION_REASON = "Action cancelled by Response Management";
-  public static final String ENABLED = "ENABLED";
-  public static final String PENDING = "PENDING";
   public static final String ACTIVE = "ACTIVE";
   public static final String CREATED = "CREATED";
-
-  @Autowired private CaseSvcClientService caseSvcClientService;
+  public static final String ENABLED = "ENABLED";
+  public static final String NOTIFY = "Notify";
+  public static final String PENDING = "PENDING";
```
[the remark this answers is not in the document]
andrewmil: Not for party statuses I don't think

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
andrewmil: Had weird issues in tests when I set some of these to final, I'll check that this is one of those cases

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -18,109 +17,143 @@
...
-          break;
-        default:
-          log.with("notification_type", notification.getNotificationType())
-              .warn("Unknown Case lifecycle event");
-          break;
-      }
-    } else {
-      log.with("action_plan_id", actionPlanIdStr)
-          .warn("Cannot accept CaseNotification for non-existent actionplan");
+    actionCaseRepo.flush();
```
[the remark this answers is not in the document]
andrewmil: In the other methods we have `...Repo.save()` or `...Repo.delete()` called before this
This `flush` is just commiting those changes? or should that happen automatically since this method has the `Transactional` decorator?

src/main/java/uk/gov/ons/ctp/response/action/service/decorator/PartyAndContact.java
```diff
@@ -18,60 +18,57 @@
 
 public class PartyAndContact implements ActionRequestDecorator {
   private static final Logger log = LoggerFactory.getLogger(PartyAndContact.class);
+  private static final String NOTIFY = "Notify";
```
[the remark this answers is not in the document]
andrewmil: :+1: to this idea but in another PR

src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
```diff
@@ -64,7 +64,12 @@ public void processActionRequests(final Action action) throws CTPException {
 
     final ActionType actionType = action.getActionType();
     if (!valid(actionType)) {
-      log.with("action_id", action.getId()).error("ActionType is not defined for action");
+      if (actionType == null) {
+        action.setActionType(ActionType.builder().name("null").build());
```
NickGrantONS: this looks dodgy

src/test/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributorTest.java
```diff
@@ -97,7 +95,7 @@ public void setUp() throws Exception {
 
     lock = mock(RLock.class);
     when(redissonClient.getFairLock(any())).thenReturn(lock);
-    when(lock.tryLock(anyLong(), anyLong(), any(TimeUnit.class))).thenReturn(true);
+    when(lock.tryLock(any(Integer.class), eq(TimeUnit.SECONDS))).thenReturn(true);
```
NickGrantONS: could use anyInt

scripts/migration-replace-bi-cases-with-b/temp_cases.csv
```diff
@@ -0,0 +1,2 @@
+83cd9eb3-af30-4993-a4e3-598fd96f4efe,35b99a90-9ad0-45b1-bfeb-f2110b402181
```
NickGrantONS: this shouldn't be in source control I guess

src/main/java/uk/gov/ons/ctp/response/action/client/PartySvcClientService.java
```diff
@@ -82,7 +71,7 @@ public PartyDTO getPartyWithAssociationsFilteredBySurvey(
         .debug("Retrieving party");
 
     final MultiValueMap<String, String> queryParams = new LinkedMultiValueMap<>();
-    queryParams.put("survey_id", Arrays.asList(surveyId));
+    queryParams.put("survey_id", Collections.singletonList(surveyId));
```
NickGrantONS: why isn't this just the surveyId? There's only ever going to be one surveyId!

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,101 +21,102 @@
...
-
-    for (final ActionType actionType : actionTypes) {
-      final List<InstructionCount> instructionCounts = processActionType(actionType);
-      distInfo.getInstructionCounts().addAll(instructionCounts);
-    }
-
-    log.debug("ActionDistributor going back to sleep");
-    return distInfo;
+  public void distribute() {
+    List<ActionType> actionTypes = actionTypeRepo.findAll();
```
NickGrantONS: cheeky little improvement - a performance enhancement. If you create a wrapper service for calling this and mark the function with`@Cachable` then Spring knows that the ActionTypes don't change very often, so it'll give you a copy from an in-memory cache. I think there's an option which specifies how long the cache copy can live. Probably out of scope for this PR, but something worth bearing in mind if you're ever doing another Spring Boot app.

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -22,101 +21,102 @@
...
+    }
+
+    try {
+      if (action.getState().equals(ActionState.SUBMITTED)) {
+        ap.processActionRequests(action);
+      } else if (action.getState().equals(ActionState.CANCEL_SUBMITTED)) {
+        ap.processActionCancel(action);
       }
+    } catch (Exception ex) {
+      // We intentionally catch all exceptions here.
```
NickGrantONS: perhaps we should put this all the way around the method contents, if it really is an intentional catch-all which would have been one level higher, but can't be because the use of lambda/foreach/`::` convenience stuff.

src/main/java/uk/gov/ons/ctp/response/action/scheduled/plan/PlanScheduler.java
```diff
@@ -15,30 +12,23 @@
  * details from the AppConfig
  */
 @Component
-public class PlanScheduler implements HealthIndicator {
+public class PlanScheduler {
```
NickGrantONS: I'm not seeing a lot of point in this class anymore

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,87 +67,54 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
-      return null;
-    }
-
-    if (actionCaseRepo.countByActionPlanFK(actionPlanPK) == 0) {
-      log.with("action_plan_pk", actionPlanPK).debug("No open cases for action plan");
+  @Transactional
+  public void createAndExecuteAllActionPlanJobs() {
+    List<ActionPlan> actionPlans = actionPlanRepo.findAll();
+    actionPlans.forEach(
+        actionPlan -> {
```
NickGrantONS: why not combine both those methods into one and use the `::` thingy?

src/main/java/uk/gov/ons/ctp/response/action/service/ActionPlanJobService.java
```diff
@@ -60,87 +67,54 @@ public ActionPlanJob findActionPlanJob(final UUID actionPlanJobId) {
...
+        });
+  }
 
-      return null;
-    }
+  private boolean hasActionableCases(ActionPlan actionPlan) {
+    return actionCaseRepo.existsByActionPlanFK(actionPlan.getActionPlanPK());
+  }
 
+  @Transactional(propagation = REQUIRES_NEW)
```
NickGrantONS: this `@Transactional` annotation is ignored when the method is called internally, so it's not creating a new transaction at all for this unit of work

src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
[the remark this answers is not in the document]
NickGrantONS: The changes will be automatically flushed to the DB before the transaction is committed, but I guess there's no harm having this here. The convention everywhere else I've seen has been to flush the changes immediately

src/main/java/uk/gov/ons/ctp/response/action/client/PartySvcClientService.java
andrewmil: I'm assuming it's because of what format the party endpoint is expecting it to be sent in but I'll test if things work if it's not in a list

src/main/java/uk/gov/ons/ctp/response/action/service/decorator/context/BusinessActionRequestContextFactory.java
```diff
@@ -21,89 +17,54 @@
...
-      if (childParty != null) {
-        childParties.add(childParty);
-      } else {
-        log.with("party_id", id).info("Unable to get party with id");
-      }
-    }
-    return childParties;
+    return respondentPartyIds
+        .stream()
+        .map(id -> partySvcClientService.getParty(SampleUnitType.BI.toString(), id))
```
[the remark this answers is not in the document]
andrewmil: Right now when you retrieve businesses/respondents from party service you define which one it is with either B or BI sample type. This doesn't really make any sense and there's a card on the board to fix this once other remove BI work is finished

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
andrewmil: I had a look at doing this but I found that it made the `processActionType` method look a bit overcomplicated. There's probably some more refactoring that could be done around here generally but might be worth saving for another PR

src/main/java/uk/gov/ons/ctp/response/action/service/ActionService.java
```diff
@@ -169,32 +181,81 @@ public Action createAction(final Action action) {
...
+      createAction(actionCase, actionRule);
+    }
+  }
 
-    final ActionPlan actionPlan =
-        actionPlanRepository.findByActionPlanPK(actionPlanJob.getActionPlanFK());
-    final Timestamp currentTime = new Timestamp((new Date()).getTime());
+  private boolean hasRuleTriggered(ActionRule rule) {
+    final Timestamp currentTime = nowUTC();
+    final Timestamp dayBefore =
```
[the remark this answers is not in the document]
andrewmil: I think we're happy with this check being in last 24 hours but we should check with business realistically

PR #80, opened 2018-08-29
Install pgcrypto extension in the public schema

johntopleyons:
# Motivation and Context
Whilst porting RM to Kubernetes running on Google Kubernetes Engine (GKE), I ran into problems with the database migrations and the installation of the **pgcrypto** PostgreSQL extension. Both the Action and Sample services use this extension and contain database migrations to install it. The problem is that as each service currently tries to install the extension in its own service schema it isn't accessible from the other service. The correct solution is to install the extension in the **public** schema, which is accessible everywhere by default.

# What has changed
* Create the extension in **schema.sql** (explicitly specifying the `public` schema) before setting the search path
* Explicitly specify the `public` schema in the extension creation statement in **schema-cloud.sql**
* Explicitly specify the `public` schema in the extension creation statement in **groundzero.sql**
* Explicitly specify the `public` schema in the extension creation statement in **add_action_rule_uuid_column.sql**

# How to test?
I created new Action and Sample service Docker images tagged _census_ and was able to test them successfully creating new pods in GKE. I have **not** tested on Cloud Foundry, nor have I tested what happens if the extension is already installed in the Action or Sample service schemas. To do so would have been far beyond my scope of getting RM working from scratch on Kubernetes.

# Links


NickGrantONS:
We are thinking about closing this PR, because it's quite old and the Travis build was problematic. Can we either get it working or close it?

johntopleyons:
Are you planning on addressing the problem, or does it just go away anyway because there isn't a Sample service for census?

NickGrantONS:
Hey @johntopleyons I think we should do this, because GC/K8s is on the radar for SDC, but it's failing the Travis build - any idea why?

NickGrantONS:
Branch is building fine, but we're parking this PR, because we don't have the bandwidth to review and release at the moment. We've raised a trello card to do this when it becomes more of a priority: https://trello.com/c/kehxzUvS/749-action-service-future-proof-for-kubernates-google-cloud

PR #91, opened 2018-09-20
Check travis fails

tej99:
# Motivation and Context

Checking travis build fails with failing test

# What has changed

# How to test?

# Links


tej99:
Travis script does not need changing, it's working as expected.

PR #96, opened 2018-09-26
Removed uneeded logic when sending INPROGRESS or NOTSTARTED cases to …

hbrace:
…FWMT

# Motivation and Context

Logic was put in to send cases to FWMT where appropriate but it was un-necessary as the action service would have handled it already.

# What has changed
Amendment to processActionRequest to remove added logic. 

# How to test?
Run ActionEndpointIT and the ActionProcessingServiceTest

# Links
[https://github.com/ONSdigital/rm-action-service/pull/92](https://github.com/ONSdigital/rm-action-service/pull/92)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
@@ -93,20 +93,14 @@ public abstract class ActionProcessingService {
       final ActionRequest actionRequest = prepareActionRequest(action);
 
       if (actionRequest != null) {
-        if (actionType.getHandler().equals("Field")) {
-          if (actionRequest.getCaseGroupStatus().equals("NOTSTARTED")
-              || actionRequest.getCaseGroupStatus().equals("INPROGRESS")) {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
```diff
@@ -93,20 +93,14 @@ public void processActionRequest(final Action action) throws CTPException {
...
-        // advise casesvc to create a corresponding caseevent for our action
-        caseSvcClientService.createNewCaseEvent(action, CategoryDTO.CategoryName.ACTION_CREATED);
-      } else {
-        log.with("action_id", action.getId())
-            .error("Unexpected situation. actionType is not defined for action");
+        actionInstructionPublisher.sendActionInstruction(actionType.getHandler(), actionRequest);
       }
+
+      // advise casesvc to create a corresponding caseevent for our action
+      caseSvcClientService.createNewCaseEvent(action, CategoryDTO.CategoryName.ACTION_CREATED);
```
andrewmil: Actions are not created at this point so this feels like the wrong place to send this case event. 
The action processing service creates action requests from already existing actions.

PR #99, opened 2018-10-08
Add REQUIRES_NEW to transaction for action distribution

andrewmil:
# Motivation and Context
Add `REQUIRES_NEW` to transaction for action distribution so actions can be processed individually

# What has changed
Added `propagation = REQUIRES_NEW` to `processActionRequests` method

# How to test?
Actions should be committed to the database individually rather than in batches on in ActionDistributor
Unit/Acceptance tests should pass

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/DistributionScheduler.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/DistributionScheduler.java
@@ -24,6 +24,10 @@ public class DistributionScheduler {
   /** Scheduled execution of the Action Distributor */
   @Scheduled(fixedDelayString = "#{appConfig.actionDistribution.delayMilliSeconds}")
   public void run() {
-    actionDistributor.distribute();
+    try {
+      actionDistributor.distribute();
+    } catch (Exception ex) {
+      log.error("Uncaught exception", ex);
+    }
   }
 }
--- a/src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
@@ -58,7 +58,7 @@ public abstract class ActionProcessingService {
    *
    * @param action the action to deal with
    */
-  @Transactional
+  @Transactional(propagation = Propagation.REQUIRES_NEW)
   public void processActionRequests(final Action action) throws CTPException {
     log.with("action_id", action.getId()).debug("Processing actionRequest");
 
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
```diff
@@ -58,7 +58,7 @@ public ActionProcessingService(ActionRequestDecorator[] decorators) {
    *
    * @param action the action to deal with
    */
-  @Transactional
+  @Transactional(propagation = Propagation.REQUIRES_NEW)
```
NickGrantONS: I suspect that the Action passed to this method, because it was from a _different_ transactions scope, could cause problems... but I shall test

andrewmil: Removing the unneeded transaction from higher up fixed this

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/DistributionScheduler.java
```diff
@@ -24,6 +24,10 @@ public DistributionScheduler(ActionDistributor actionDistributor) {
   /** Scheduled execution of the Action Distributor */
   @Scheduled(fixedDelayString = "#{appConfig.actionDistribution.delayMilliSeconds}")
   public void run() {
-    actionDistributor.distribute();
+    try {
+      actionDistributor.distribute();
+    } catch (Exception ex) {
```
mfcrocker: Having had a quick look I think the answer is no, but can we be more specific than `Exception` here?

src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
```diff
@@ -167,10 +167,15 @@ private ActionCancel prepareActionCancel(final Action action) {
...
-      throws CTPException {
-    final ActionDTO.ActionState nextState =
-        actionSvcStateTransitionManager.transition(action.getState(), event);
+  private void transitionAction(final Action action, final ActionDTO.ActionEvent event) {
+    ActionDTO.ActionState nextState = null;
+
+    try {
+      nextState = actionSvcStateTransitionManager.transition(action.getState(), event);
+    } catch (CTPException ctpExeption) {
+      throw new IllegalStateException();
```
mfcrocker: I'd prefer if aSSTM.transition() was changed to throw `IllegalStateException` rather than catching `CTPException` and throwing a more specific one

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/DistributionScheduler.java
NickGrantONS: Reason for catch-all is that we want to know what kinds of runtime exceptions and things are bubbling up all the way to the top here.

src/main/java/uk/gov/ons/ctp/response/action/service/ActionProcessingService.java
NickGrantONS: yes I very much agree. It's one big refactor I want to do, to eliminate CTPException from the whole project - all 7+ Java 'micro' services do not need this checked exception and should not be using it like this.

PR #103, opened 2018-10-15
Year added to return by date for FWMT

hbrace:
# Motivation and Context
FWMT tool requires the return by date sent to them to include the year.

# What has changed
SOCIALICF added to clause where it decides which date format to use. A unittest has been added to test that the correct date is used, and the existing social and business unittests have had an action type added to ensure they have the correct date formats.

# How to test?
This can be tested by creating two collection exercises in the rasrm-ops tool locally, one for SOCIALICF and one for SOCIALNOT. Once these are ran, you should see that the messages on the Field RabbitMQ queue have the return by date in the format dd/mm/yyyy, and the messages on the Printer queue should have the return by date in the format dd/mm. 

# Links
https://trello.com/c/kvBHikuB


hbrace:
[BUG302: Send 'return by' year to FWMT](https://trello.com/c/kvBHikuB/302-send-return-by-year-to-fwmt)

PR #105, opened 2018-10-22
Bug fix include pending enrolments in print file

gemmai95:
# Motivation and Context

Currently pending enrolments are missing from the print file this change should fix that.

# What has changed

Change query parameter sent of enrolment statuses to party from a list to comma separated string. 

# How to test?

When running an action rule with a business that only has a pending enrolment for a survey they party service is called to get respondent details of this pending enrolment.

# Links

Trello: https://trello.com/c/Cwkh7tWn

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/client/PartySvcClientService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/client/PartySvcClientService.java
@@ -78,9 +78,22 @@ public class PartySvcClientService {
         .with("survey_id", surveyId)
         .debug("Retrieving party");
 
+    StringBuilder commaSepEnrolmentStatuses = new StringBuilder();
+
+    for (int i = 0; i < enrolmentStatuses.size(); i++) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/ons/ctp/response/action/client/PartySvcClientService.java
```diff
@@ -78,9 +78,14 @@ public PartyDTO getPartyWithAssociationsFilteredBySurvey(
...
 
+    String commaSepEnrolmentStatuses = String.join(", ", enrolmentStatuses);
+
     final MultiValueMap<String, String> queryParams = new LinkedMultiValueMap<>();
     queryParams.put("survey_id", Collections.singletonList(surveyId));
-    queryParams.put("enrolment_status", enrolmentStatuses);
+    queryParams.put(
+        "enrolment_status", Collections.singletonList(commaSepEnrolmentStatuses.toString()));
+
+    log.with(queryParams).info("QueryParams");
```
andrewmil: Was this log line just for debugging?

PR #106, opened 2018-10-23
Set action state to aborted if case does not exist

gemmai95:
# Motivation and Context

Abort actions with missing cases. Fixes actions getting stuck in a submitted state when someone enrols and completes within a certain time frame.

# What has changed

Instead of throwing an exception when an action case is missing abort the action.

# How to test?

# Links

Trello: https://trello.com/c/EKVubdKl

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
@@ -129,16 +129,18 @@ class ActionDistributor {
     ActionCase actionCase = actionCaseRepo.findById(action.getCaseId());
 
     if (actionCase == null) {
+      log.with("action", action).info("Case no longer exists for action");
+      ActionState newActionState;
       try {
-        log.with("action", action).info("Case no longer exists for action");
-        ActionState newActionState =
+        newActionState =
             actionSvcStateTransitionManager.transition(
                 action.getState(), ActionEvent.REQUEST_CANCELLED);
-        action.setState(newActionState);
-        actionRepo.saveAndFlush(action);
-      } catch (Exception ex) {
+      } catch (CTPException ex) {
         throw new IllegalStateException();
       }
+
+      action.setState(newActionState);
+      actionRepo.saveAndFlush(action);
     }
 
     SampleUnitDTO.SampleUnitType caseType =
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -122,8 +122,8 @@ private ActionProcessingService getActionProcessingService(Action action) {
     ActionCase actionCase = actionCaseRepo.findById(action.getCaseId());
 
     if (actionCase == null) {
-      log.with("action", action).error("Cannot find case for action");
-      throw new IllegalStateException();
+      log.with("action", action).info("Case no longer exists for action");
+      action.setState(ActionState.ABORTED);
```
NickGrantONS: This bypasses the state transition stuff. Are we OK to do that?

NickGrantONS: also, there's no save to persist the change to the DB

gemmai95: Good shout completely spaced on that. Changing to work with transitions

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -118,12 +125,16 @@ private void processAction(Action action) {
     }
   }
 
-  private ActionProcessingService getActionProcessingService(Action action) {
+  private ActionProcessingService getActionProcessingService(Action action) throws CTPException {
```
NickGrantONS: rather than declaring `throws CTPException` I'd prefer it if we caught the exception and threw an `IllegalStateException` which is a runtime exception, so it doesn't need to be 'checked'. It's a few more lines of code, but it's a step towards getting rid of checked exceptions

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -45,6 +49,9 @@
   private ActionProcessingService businessActionProcessingService;
   private ActionProcessingService socialActionProcessingService;
 
+  @Autowired
+  private StateTransitionManager<ActionState, ActionEvent> actionSvcStateTransitionManager;
```
andrewmil: For consistency this should be injected through the constructor

src/main/java/uk/gov/ons/ctp/response/action/scheduled/distribution/ActionDistributor.java
```diff
@@ -122,8 +129,18 @@ private ActionProcessingService getActionProcessingService(Action action) {
     ActionCase actionCase = actionCaseRepo.findById(action.getCaseId());
 
     if (actionCase == null) {
-      log.with("action", action).error("Cannot find case for action");
-      throw new IllegalStateException();
+      log.with("action", action).info("Case no longer exists for action");
```
andrewmil: Untested

PR #108, opened 2018-10-24
416 remove bi references

AndrewTorrance:
# Motivation and Context
Changes due to the party get business by if and by reference dropping the sample unit type 

# What has changed
 Changed end point definition in yml files and stopped sending sample unit type 

# How to test?
Acceptance tests

# Links
https://trello.com/c/I5lc4gDq


AndrewTorrance:
The is PR has become stale and relies on BI cases being removed which may lead to further work. Hence closing this with a view to reappraising AFTER BI cases removed.

PR #110, opened 2018-10-24
Sus011 Send updates to FWMT when case reaches an in-actionable status

rjweeks:
# **Send updates to FWMT when case reaches an in-actionable status**

# Motivation and Context
Send Cancellation of SocialICF reminders to Field Management. 

# Links
https://trello.com/c/T7fTfnvz/253-sus011-send-updates-to-fwmt-when-case-reaches-an-in-actionable-status-5


src/main/java/uk/gov/ons/ctp/response/action/service/CaseNotificationService.java
```diff
@@ -71,6 +75,7 @@ public void acceptNotification(final CaseNotification notification) throws CTPEx
       case DISABLED:
       case DEACTIVATED:
         actionService.cancelActions(caseId);
+        socialActionProcessingService.cancelFieldWorkReminder(caseId);
```
davidmort: we need to be careful that we only do this for 'H' case types

src/main/java/uk/gov/ons/ctp/response/action/service/SocialActionProcessingService.java
```diff
@@ -0,0 +1,90 @@
...
+import uk.gov.ons.ctp.response.action.service.decorator.CollectionExerciseAndSurvey;
+import uk.gov.ons.ctp.response.action.service.decorator.SampleAttributes;
+import uk.gov.ons.ctp.response.action.service.decorator.context.ActionRequestContextFactory;
+import uk.gov.ons.ctp.response.casesvc.representation.CaseDetailsDTO;
+
+@Service
+@Qualifier("social")
+public class SocialActionProcessingService extends ActionProcessingService {
+
+  private static final Logger log = LoggerFactory.getLogger(SocialActionProcessingService.class);
```
liamtoozer: Does this class need any logging? If not, can the logger be removed here?

rjweeks: Good spot. I will add logging for the cancel event.

rjweeks:
[SUS011: Send updates to FWMT when case reaches an in-actionable status (5)](https://trello.com/c/T7fTfnvz/253-sus011-send-updates-to-fwmt-when-case-reaches-an-in-actionable-status-5)

PR #113, opened 2018-11-02
Only send cancellation if SOCIAL ICF action is found.

rjweeks:
# Motivation and Context
If no SOCIAL ICF action found it throws a null pointer exception.


rjweeks:
[Fix Null pointer exception for case in-actionable notifications.](https://trello.com/c/t5hscn1v/341-fix-null-pointer-exception-for-case-in-actionable-notifications)

PR #120, opened 2019-07-26
Add a 1 minute cache evict to collex

mfcrocker:
# Motivation and Context
We were having issues with an infinite cache on collection exercise details in this service. This evicts the cache once a minute

# What has changed
Added a `CacheEvict` method on a 6000ms schedule

# How to test?
Not super easy to test really :(

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/client/CollectionExerciseClientService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/client/CollectionExerciseClientService.java
@@ -74,6 +74,6 @@ public class CollectionExerciseClientService {
   }
 
   @CacheEvict(cacheNames = "collectionExercise", allEntries = true)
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/ons/ctp/response/action/client/CollectionExerciseClientService.java
```diff
@@ -70,4 +72,8 @@ public CollectionExerciseDTO getCollectionExercise(final UUID collectionExcercis
     log.with("collection_exercise", result).debug("made call to collection Exercise");
     return result;
   }
+
+  @CacheEvict(cacheNames = "collectionExercise", allEntries = true)
+  @Scheduled(fixedRate = 6000)
```
AndrewTorrance: I think this will clear the cache every 6 seconds ?

PR #122, opened 2019-09-26
Improve exception handling

insacuri:
# Motivation and Context
Originally this PR was to fix the error handling in this service.  Turns out that's a lot more involved then expected and can cause things to happen that we're not expecting.
So instead, this PR is just going to be a tidy up of the logging by moving certain things to info level so we get better visibility of them.  It also adds more metadata to the logging where possible to make debugging issues easier.

Finally there was an error line that is less an error and more of a warning (as it happens occasionally but nothing falls over which isn't great, but errors should be for when things are going to be on fire)

# What has changed
- A few log level changes
- More metadata in certain logging

# How to test?
- mvn clean test

# Links


insacuri:
Merged it despite failing build as there's nothing wrong with the code.  Something is up between travis and the artifactory

PR #123, opened 2019-10-01
Up CSV ingester polling interval to an hour

mfcrocker:
# Motivation and Context
The CSV ingester aggressively polls for content and it's unnecessary. This ups it from 30 seconds to an hour


mfcrocker:
@insacuri What is Java? We just don't know

PR #137, opened 2020-03-17
added non-root user to dockerfile

anwilkie:
# Motivation and Context
The Dockerfile was running as the default root user and was failing to boot. A non-root user needed to be created with privileges.

# What has changed

- Modified `_infra/docker/Dockerfile` to create a new user (docker_user), add the user to the docker group (to give it root privileges), and set the new user as the current one.

# How to test?
Make sure the Dockerfile works.

# Links

[Trello card](https://trello.com/c/UMhnP9sR)

### DIFF
--- a/_infra/docker/Dockerfile
+++ b/_infra/docker/Dockerfile
@@ -1,10 +1,11 @@
 FROM openjdk:8-jre-slim
 
-RUN useradd docker_user && \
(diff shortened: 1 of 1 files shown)


mfcrocker:
If this is still required, please rebase and reopen it

PR #146, opened 2020-05-06
Removing zipkin

anwilkie:
# Motivation and Context
Zipkin needed to be removed from the various services.

# What has changed
Removed all mentions of Zipkin.

# Links
[Trello card](https://trello.com/c/YZ9aio6I)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,6 +129,12 @@
             <artifactId>spring-rabbit</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-crypto</artifactId>
+            <version>4.2.3.RELEASE</version>
+        </dependency>
+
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -124,15 +124,15 @@
             <artifactId>spring-cloud-spring-service-connector</artifactId>
         </dependency>
 
-        <!-- https://mvnrepository.com/artifact/org.springframework.cloud/spring-cloud-sleuth-zipkin -->
         <dependency>
-            <groupId>org.springframework.cloud</groupId>
-            <artifactId>spring-cloud-sleuth-zipkin</artifactId>
+            <groupId>org.springframework.amqp</groupId>
```
jterryons, line 128: is this required? I would have thought this would already be a part of spring-cloud-starter

anwilkie: Adding this dependency fixed the travis build error. (Check the "added spring security dependency" commit)

PR #159, opened 2020-06-24
Document action service

mfcrocker:
Documentation of the action service. Suggestions on Trello, but TLDR: this service needs merging with the ones it calls all the time. Merging with Case should be seen as non-optional.


API.yaml
```diff
@@ -0,0 +1,949 @@
...
+        - action-rules
+      operationId: deleteActionRuleUsingDELETE
+      parameters:
+        - name: actionRuleId
+          in: path
+          description: actionRuleId
+          required: true
+          schema:
+            type: string
+      requestBody:
```
jterryons, line 366: I thought you might have added this in error but then i checked the code.
wake me when its over.

mfcrocker: Yeah I mean, it sure violates principles to the degree that OpenAPI will moan about it

mfcrocker: Yeah I mean, it sure violates principles. So badly in fact that OpenAPI will moan about it

PR #165, opened 2020-09-07
Integrate pubsub

jterryons:
Send an action request over pubsub. An action request is a do everything god object. We do not have access to the templates Ids anymore and the old notify service attempted to get the template the from comms-template service via the classifiers. Fortunately the new notify service is able to retrieve the template itself when classifiers are supplied. A new NotifyModel object has been created to supply the service with a structure it can understand.

There appears to be spaces in the json keys of our personalisation objects. thats just how it was originally designed.

I noticed that maven wasnt detecting any new unit tests, updating the version of surefire fixed this but also highlighted ~200 unit tests, which presumably, were never running in our pipelines.


src/main/java/uk/gov/ons/ctp/response/action/service/NotifyService.java
```diff
@@ -0,0 +1,99 @@
...
+      ByteString data = ByteString.copyFromUtf8(message);
+      PubsubMessage pubsubMessage = PubsubMessage.newBuilder().setData(data).build();
+
+      ApiFuture<String> messageIdFuture = publisher.publish(pubsubMessage);
+      String messageId = messageIdFuture.get();
+      log.with("messageId", messageId)
+          .with("actionId", actionRequest.getActionId())
+          .debug("Notify pubsub sent sucessfully");
+
+      // ActionFeedback feedback = new ActionFeedback();
```
insacuri, line 45: If the commented out block isn't needed then it should be deleted

src/main/java/uk/gov/ons/ctp/response/action/service/NotifyService.java
```diff
@@ -0,0 +1,99 @@
...
+
+    return new NotifyModel(
+        Notify.builder()
+            .personalisation(personalisation)
+            .classifiers(classifiers)
+            .emailAddress(actionRequest.getContact().getEmailAddress())
+            .build());
+  }
+
+  // public Classifiers getClassifiers(final ActionRequest actionRequest) {
```
insacuri, line 95: Delete commented out code

warrenbailey (approved):

PR #169, opened 2020-09-29
Removing the action feedback

warrenbailey:
# Motivation and Context

Removing the action feedback path as this is only used by the notify gateway (now retired) and the action exporter (approaching retirement)

# What has changed

# How to test?

# Links


jdfox (approved):

insacuri (approved):
Not tested it, but I'm almost certain it's not used and shouldn't cause any issues.  LGTM

PR #180, opened 2020-10-09
Remove feedback

insacuri:
# Motivation and Context
Removes all the code relating to feedback as it's something we don't use anymore.

# What has changed

# How to test?

# Links


insacuri:
Tested it locally and it seemed fine, but without an EQ to complete the process, I can't be sure of the whole journey

warrenbailey (approved):

warrenbailey:
Make sense to test on preprod then

PR #181, opened 2020-10-14
Move Action Exporter into Action

warrenbailey:
This is a straight port of the remaining code in Action Exporter (https://github.com/ONSdigital/rm-actionexporter-service)

Both "applications" now exist inside the same spring context and communicate via a method call rather than a rabbit queue.

The next stage will be to tightly integrate the two of them.


src/main/java/uk/gov/ons/ctp/response/action/message/ActionInstructionPublisher.java
```diff
@@ -36,7 +39,11 @@ public void sendActionInstruction(@Header("HANDLER") final String handler, final
       instruction.setActionCancel((ActionCancel) action);
     }
 
-    final String routingKey = String.format("%s%s%s", ACTION, handler, BINDING);
-    rabbitTemplate.convertAndSend(routingKey, instruction);
+    // if (ActionRequestValidator.ACTIONEXPORTER.equals(handler)) {
```
insacuri, line 42: Remove commented out code

jterryons (approved):

_infra/helm/action/values.yaml
```diff
@@ -38,6 +43,14 @@ crons:
     name: action-plan-executor
     cron: "* * * * *"
     target: "actionplans/execute"
+  recordDeleter:
+    name: old-record-deleter
```
insacuri, line 47: Would it be an easy change to rename both the cronjobs to `action-old-record-deleter` and `action-export-scheduler`?

If not, we can do it as another bit of work down the line

src/main/java/uk/gov/ons/ctp/response/action/message/ActionInstructionPublisher.java
```diff
@@ -36,6 +40,11 @@ public void sendActionInstruction(@Header("HANDLER") final String handler, final
       instruction.setActionCancel((ActionCancel) action);
     }
 
+    if (ActionRequestValidator.ACTIONEXPORTER.equalsIgnoreCase(handler)) {
+      actionExportService.acceptInstruction(instruction);
+    }
+
+    // TODO we should be able to remote this soon
```
insacuri, line 47: remote?  Do you mean remove?
Also, maybe put more context as to when it can be removed i.e., 'can be removed once we've switched to using pubsub fully'

insacuri (approved):
Outside those 2 minor comments, the changes look good

src/main/java/uk/gov/ons/ctp/response/action/message/ActionInstructionPublisher.java
warrenbailey: I think this can be removed now

_infra/helm/action/values.yaml
warrenbailey: Yeah np

PR #187, opened 2020-11-05
Remove unused code

insacuri:
# Motivation and Context

# What has changed

# How to test?

# Links


src/main/java/uk/gov/ons/ctp/response/lib/common/time/DateTimeUtil.java
```diff
@@ -1,25 +1,12 @@
 package uk.gov.ons.ctp.response.lib.common.time;
```
akssinhaamit, line 1: Is this not being used by schedulers?

src/main/resources/application.yml
```diff
@@ -88,18 +88,6 @@ data-grid:
 
   report-lock-time-to-live-seconds: 300
 
-# the thread that sends actions to handlers
```
akssinhaamit, line 91: Is it not being used by scheduler?

akssinhaamit (changes requested):
`ActionRequestValidator.java` can be removed I believe.

src/main/resources/application.yml
insacuri: That was used by the old version of the scheduler (with redis and the @scheduled annotation).  The new one has its timing set in the helm chart

src/main/java/uk/gov/ons/ctp/response/lib/common/time/DateTimeUtil.java
insacuri: I couldn't find anything in the codebase that was using the 2 functions I deleted.  Can you see anywhere that it's used?

akssinhaamit (approved):

PR #188, opened 2020-11-09
Shutdown publisher after using it

warrenbailey:
# Motivation and Context

# What has changed

# How to test?

# Links


mfcrocker (approved):
mvn fmt else fine

PR #211, opened 2020-12-22
Adding default to active enrolment

akssinhaamit:
# Motivation and Context
Setting active enrolment default to false so that old records not break.

# What has changed

# How to test?

# Links


warrenbailey (approved):

warrenbailey:
AS an FYI this PR was merged and then reverted https://github.com/ONSdigital/rm-action-service/pull/207/

That also used release 22 as the next liquibase script so unfortunately preprod thinks that is already applied.

PR #214, opened 2021-01-08
update active_enrolment column for all existing record in case table …

akssinhaamit:
# Motivation and Context
 update active_enrolment column for all existing record in case table with associated action plan

# What has changed
The script has been added. Its not added to the change log which will be updated once we have a heads up for action svc deprecation.
# How to test?

# Links

### DIFF
--- a/src/main/resources/database/changelog-master.yml
+++ b/src/main/resources/database/changelog-master.yml
@@ -77,4 +77,4 @@ databaseChangeLog:
 
 # this should only be deployed to prod once we are ready to turn on the actionsvc deprecated to true
 - include:
-      file: database/changes/release-24/changelog.yml
+      file: database/changes/release-25/changelog.yml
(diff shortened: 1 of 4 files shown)


src/main/resources/database/changelog-master.yml
```diff
@@ -73,4 +73,8 @@ databaseChangeLog:
       file: database/changes/release-23/changelog.yml
 
 - include:
-    file: database/changes/release-24/changelog.yml
\ No newline at end of file
+    file: database/changes/release-24/changelog.yml
+
+# this should only be deployed to prod once we are ready to turn on the actionsvc deprecated to true
+- include:
+      file: database/changes/release-24/changelog.yml
```
warrenbailey, line 80: should this be release-25?

warrenbailey (changes requested):

src/main/resources/database/changelog-master.yml
akssinhaamit: my bad

warrenbailey (approved):

PR #216, opened 2021-01-20
Big sample retry logic

akssinhaamit:
# Motivation and Context
Big Sample files creates a scenario wherein the sample cases is still being propagated to action service and the action event time lapses. Due to which we miss data

# What has changed
A new feature has been provided which makes action service self sufficient to manage any cases which is still in flight.

# How to test?
Smoke test with a big sample file

# Links
[Trello](https://trello.com/c/3ZKkA66f/419-s26-issue-with-big-sample-cases-not-propagating-to-actions)


src/main/java/uk/gov/ons/ctp/response/action/service/ProcessEventService.java
```diff
@@ -135,6 +141,44 @@ public void processEvents(UUID collectionExerciseId, String eventTag) {
...
+  }
+
+  private void partialProcessCheckPoint(
+      UUID collectionExerciseId,
+      String eventTag,
+      Optional<ActionEventPartialEntry> partialEntry,
+      Instant instant) {
+    Long numberOfActionCases =
+        actionCaseRepository.findByCollectionExerciseId(collectionExerciseId);
+    Long numberOfCases = caseSvcClientService.getNumberOfCases(collectionExerciseId);
```
warrenbailey, line 154: The getNumberOfCases method can return Null if it errors, but you're not checking for that here so you risk a null pointer if case service is down.

src/main/java/uk/gov/ons/ctp/response/action/service/ProcessEventService.java
```diff
@@ -220,7 +264,7 @@ private boolean processLetterCases(
    * @param status
    * @param instant
    */
-  private void createCaseActionEvent(
+  public void createCaseActionEvent(
```
warrenbailey, line 267: Any reason why you've changed these methods from private to public? As far as I can see they're only used by this class?

warrenbailey (changes requested):

src/main/java/uk/gov/ons/ctp/response/action/service/ProcessEventService.java
akssinhaamit: 👍

src/main/java/uk/gov/ons/ctp/response/action/domain/repository/ActionCaseRepository.java
```diff
@@ -38,4 +40,9 @@
       UUID collectionExerciseId, boolean activeEnrolment);
 
   List<ActionCase> findByIdIn(List<UUID> caseId);
+
+  @Query(
+      "SELECT count (*) FROM ActionCase c "
+          + "WHERE c.collectionExerciseId = :collectionExerciseId")
+  Long findByCollectionExerciseId(@Param("collectionExerciseId") UUID collectionExerciseId);
```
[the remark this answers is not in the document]
akssinhaamit: 👍

src/main/java/uk/gov/ons/ctp/response/action/service/ProcessEventService.java
akssinhaamit: I started doing refactoring and changed the access identifier. reverted

warrenbailey (approved):

PR #231, opened 2021-03-10
refactor action service inital commit

akssinhaamit:
# What and why?

# How to test?

# Trello


insacuri (commented):
The deletions look reasonable.  Just want to test it against my environment and it should be good to go

insacuri (approved):
Feels weird to approve such a big PR after such a small time reviewing, but I tested it in my dev environment and got MPS and reminders to send so I'm happy this should work fine!

PR #237, opened 2021-05-12
Updating schema column names

anwilkie:
# What and why?
The schema column names need to use underscores to separate words to maintain consistent formatting.

# How to test?
Check that column names have changed.

# Trello
[Card](https://trello.com/c/bwTYlY8m)


src/main/resources/database/changes/release-29/mofiying_column_names.sql
```diff
@@ -0,0 +1,23 @@
...
+RENAME COLUMN partyid TO party_id
+RENAME COLUMN collectionexerciseid TO collection_exercise_id
+RENAME COLUMN sampleunittype TO sample_unit_type
+RENAME COLUMN casepk TO case_pk;
+
+ALTER TABLE action.actionplan
+RENAME COLUMN actionplanpk TO action_plan_pk
+RENAME COLUMN createdby TO created_by
+RENAME COLUMN lastrundatetime TO last_run_date_time;
+
```
warrenbailey, line 15: you can't modify the the databasechangelog tables because they are managed and used by liquibase so these will need to be reverted.

warrenbailey (changes requested):
looks good, but we can't modify the liquibase tables.

warrenbailey (approved):

PR #245, opened 2021-08-17
Pubsub spring integration test

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/126JzDkR/871-implement-spring-integration-for-pubsub-between-case-service-and-action-service


src/main/java/uk/gov/ons/ctp/response/action/message/CaseNotificationReceiver.java
```diff
@@ -3,66 +3,41 @@
...
+    try {
+      log.info("Mapping payload to CaseNotification object");
+      CaseNotification caseNotificationReceiver =
+          objectMapper.readValue(payload, CaseNotification.class);
+      log.info("Mapping successful, accepting action case notification");
+      caseNotificationService.acceptNotification(caseNotificationReceiver);
+    } catch (final IOException | CTPException e) {
+      log.with(e)
+          .error(
+              "Something went wrong while processing message received from PubSub for action case notification");
```
warrenbailey, line 40: will this stop it nacking the message? 

I think spring relies on exceptions to know whether to nack or ack the messages

warrenbailey: From here: https://cloud.spring.io/spring-cloud-gcp/multi/multi__spring_integration.html
```text
Automatic acking (AckMode.AUTO)

A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown. If a RuntimeException is thrown while the message is processed, then the message is nacked.
```

creativeappdevamit: you are right! changed it to Manual mode.

creativeappdevamit: updated

insacuri (approved):

PR #246, opened 2021-08-31
Add reference field to notify emails

mfcrocker:
# What and why?
This adds the reference field to GovNotify emails so they'll be picked up in reports download from the site

# How to test?

# Trello
https://trello.com/c/e57pqgz3/885-s42-rops-overhaul-provide-additional-details-for-govnotify-reports

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/service/ProcessEventService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/service/ProcessEventService.java
@@ -328,6 +328,7 @@ public class ProcessEventService {
     existingEvent.setProcessedTimestamp(Timestamp.from(instant));
     actionEventRepository.save(existingEvent);
   }
+
   /**
    * Populates data for print file for case event
    *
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/ons/ctp/response/action/service/ProcessEventService.java
```diff
@@ -720,6 +721,7 @@ private void processEmail(
                 .personalisation(personalisation)
                 .classifiers(classifiers)
                 .emailAddress(respondentParty.getAttributes().getEmailAddress())
+                .reference(survey.getSurveyRef() + "-" + businessParty.getAttributes().getRuref())
```
insacuri, line 724: Is it possible for either of these to be blank?  I don't think so but I figured I'd ask.

insacuri (approved):

src/main/java/uk/gov/ons/ctp/response/action/service/ProcessEventService.java
mfcrocker: Not as far as I'm aware, and if somehow either was I'm not sure how we'd even handle that

PR #247, opened 2021-08-31
spring integration issues fix

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/b/EgV9fGPY/rasrm

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/action/ActionSvcApplication.java
+++ b/src/main/java/uk/gov/ons/ctp/response/action/ActionSvcApplication.java
@@ -26,6 +26,7 @@ import org.springframework.integration.annotation.ServiceActivator;
 import org.springframework.integration.channel.DirectChannel;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.handler.annotation.Header;
 import org.springframework.scheduling.annotation.EnableAsync;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/ons/ctp/response/action/ActionSvcApplication.java
```diff
@@ -177,22 +174,10 @@ public PubSubInboundChannelAdapter actionCaseNotificationAdapter(
     PubSubInboundChannelAdapter adapter =
         new PubSubInboundChannelAdapter(pubSubTemplate, subscriptionName);
     adapter.setOutputChannel(inputChannel);
+    adapter.setAckMode(AckMode.MANUAL);
     return adapter;
   }
 
-  @Bean
-  @InboundChannelAdapter(
```
insacuri, line 184: It's still using a DirectChannel instead of PublishSubscribeChannel.   Is that  intentional?

insacuri (changes requested):

src/main/java/uk/gov/ons/ctp/response/action/ActionSvcApplication.java
creativeappdevamit: It shouldn't matter as its default. But to be consistent I will change it to PublishSubscribeChannel. Its good to have it consistent.

insacuri (approved):

